PR #85, opened 2026-05-02
v0.7 Sprint 2: performance follow-up (PERF-061..064, TCK-064, HEUR-061)

arkstack-dev:
PERF-064: replace LongAdder.reset() with baseline gating in CoreFlowRuntime to eliminate the close()/start() race when interruptAndJoinRunningThreads hits its 5s timeout. Each lifecycle counter now exposes (totalSinceForever - baselineAtTransition); baseline writes are atomic volatile longs, so a stale worker increment after close cannot leave the next generation with a non-zero residual.

PERF-063: replace ConcurrentLinkedQueue with MpscUnboundedArrayQueue in NativeTcpCarrier.ReactorLoop.pendingRequests. MPSC semantics match exactly (many enqueueRegistration/enqueueWriteInterest/cancelKey producers, single reactor consumer); chunked allocation avoids per-element node alloc churn under burst arrivals; matches the pattern already used in NativeTcpStream.outboundQueue.

HEUR-061: design-note assessment instead of refactor. NativeTcpCarrier and NativeTcpStream stay single-file; full responsibility-line decomposition deferred until io_uring carrier work or 2k LOC threshold (whichever arrives first). Documented in docs/subsystems/transport.md.

PERF-062: TLS ingress hot path in NativeTcpStream.decryptIngress and readTlsIngressFromFd. Fast-fail check before allocation when offerIngress would reject (closed stream / queue at capacity); ownership-transfer flag replaces the retain()/auto-close ceremony around try-with-resources, so the success path no longer pays a retain+close pair.

PERF-061: HTTP/2 frame writer in CommunityHttp2SessionProcessor uses a per-response reusable outbound carrier sized to MAX frame header + payload. One borrow now replaces N per-frame allocations on the (HEADERS/CONTINUATION
+ N DATA) wire-image path. Handshake helpers (PING ACK, SETTINGS, RST_STREAM, GOAWAY) keep per-call allocation since they are one-shot and rare.

TCK-064: NativeTcpTransportStressTest gains a readFully short-read loop — TransportStream.read returns whatever is available, not necessarily the full requested size. The single-shot read assertion was hypothesis (3) for the constrained-CI deadlock. Hypothesis (2) (CLQ ordering anomaly) is addressed by PERF-063 above. Hypothesis (1) (thread-pressure deadlock with 10 client engines × 1 reactor + 4 server reactors on 2 vCPUs) requires a structural refactor (single client engine + N streams) deferred to v0.8. Investigation status documented in the test class Javadoc; transport-stress-gate nightly CI job (-DincludedGroups=stress) still recommended at the ops level.

Verification:
- Core Flow tests: 45/45 green (PERF-064 baseline behavior preserved)
- Community transport/HTTP/Flow sweep: 184 tests, 0 failures, 0 errors, 2 skipped (stress + bench, default-excluded by tag)
- Stress variant under -Dgroups=stress: 4.5s green locally with PERF-063
  + short-read fix applied

### DIFF
--- a/docs/subsystems/transport.md
+++ b/docs/subsystems/transport.md
@@ -321,6 +321,47 @@ for client IP preservation behind load balancers (HAProxy, NGINX, AWS NLB, GCP L
 
 ---
 
+## Implementation Notes — Class Decomposition Assessment (HEUR-061)
+
+`NativeTcpCarrier` (~1.4k LOC, 146 declared members) and `NativeTcpStream` (~1.2k LOC, 133 declared members) cross the CLAUDE.md "≈5 collaborators" heuristic threshold. A v0.6 PR-review carry-over (HEUR-061) asked whether they should be decomposed along reactor / FD-owner / PAQS-dispatch responsibility lines.
+
+**Assessment outcome (v0.7 Sprint 2): keep current single-file shape; do not refactor in this sprint.**
+
+**Responsibility lines (informational map, not extraction targets):**
+
+`NativeTcpCarrier`:
+
+- **Engine lifecycle + config** — `TransportConfig` / `MemoryAllocator` / `KernelCryptoProvider` injection, `running` / `closed` AtomicBooleans, acceptor thread, stream/connection counters.
+- **FFM socket backend selection + validation** — `SocketBackendMode` enum, `SocketBackendSelection` record, PosixHybrid / NIO fallback, `validateServerSocketBootstrap*` / `validateClientSocketBackend*` probes, Windows `bestEffortWsaCleanup`.
+- **Reactor loop** — `ReactorLoop` inner class: `Selector`, `pendingRequests` MPSC queue (PERF-063 — `MpscUnboundedArrayQueue`), `drainPendingRequests`, key dispatch.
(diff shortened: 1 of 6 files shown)


arkstack-dev:
## Exeris Kernel — PR Review
**Triage:** HOT_PATH · TLS ingress ownership transfer, HTTP/2 frame carrier reuse, LongAdder baseline race fix, MPSC queue substitution.
**Verdict: APPROVE**

**Strong patterns**

- **PERF-062** — ownership-transfer flag (`transferred`) is a strict improvement over the old `retain()`/TWR pair: eliminates the refcount round-trip on the success path while keeping the failure and exception paths fully covered in `finally`. Both `readTlsIngressFromFd` and `decryptIngress` are correctly symmetric.
- **PERF-061** — reusing `outboundFrame` across all frames in a response is safe here: the old code closed the per-frame buffer immediately after the same `stream.write` call, which proves `write` is copy-out synchronous; a deferred write would have been broken under the old pattern too. Carrier is held in the outer TWR, closed after all frames complete — no leak path.
- **PERF-064** — volatile baseline snapshot vs. racy `LongAdder.reset()` is the correct fix. `Math.max(0L, …)` guard on `parkedFlows` only (bidirectional counter) with unguarded subtraction on the monotonically-increasing counters is the right asymmetry.

---
*False positive? Reply `@claude false-positive <id>` or 👎.*

---

PR #92, opened 2026-05-08
v0.7 Sprint 6b: Kafka empty-subscription guard + wake test re-enabled

arkstack-dev:
## Summary

Sprint 6a deferred `wakeDecision_wakesParkedFlow` under Kafka transport with a working hypothesis of a bridge/lookupParked race. **Diagnostic run disproved that hypothesis** — the wake event never reached the bridge because the consumer-loop VT crashed on its first `poll()` call with an empty subscription set.

**Root cause.** `KafkaConsumer.poll()` in kafka-clients 3.x throws `IllegalStateException` when called with neither subscriptions nor manual assignments. The race is *start-vs-register*: `engine.start()` launches the consumer loop before any caller registers an event type, and if the consumer-loop VT reaches its first `poll()` before the test thread calls `registry.register(...)`, the loop crashes silently and all subsequent dispatches are lost. Start/Ignore/RAII passed by winning the race; Wake lost it because it does `schedule + wait-for-park` between `setUp()` and `registerType()`.

**Fix.** `KafkaEventEngine.ConsumerLoop`: guard the steady-state poll path with a `subscribedTopics.isEmpty()` check that `LockSupport.parkNanos` for the same duration as `poll()`. The loop stays responsive to `register()` bumping the registry version; no busy-spin.

**Test surface.**
- Re-enabled `wakeDecision_wakesParkedFlow` in `CommunityKafkaFlowChoreographyTckIT` (removed `@Disabled` override + Sprint 6b deferral block); 30 s budget unchanged.
- New regression test on `AbstractKafkaEventEngineTck`: `startWithoutRegistrationsThenRegisterAndPublish` covers `engine.start()` before any registry entry, sleeps past one poll cycle, asserts `loop().isRunning()`, then registers + publishes and verifies delivery. Catches the bug for any future Kafka binding.

## Test plan

- [x] Unit suite — 1017/1017 green (12 skipped = integration tests by tag).
- [x] Quality gates — 0 PMD / 0 Checkstyle on touched modules (spi, tck, core, community, community-kafka).
- [x] Kafka EventEngine TCK IT — 5/5 green (including the new regression test) in ~6 s.
- [x] Kafka FlowChoreography TCK IT — 4/4 green (Wake now passes) in ~10 s.

## Out of scope (Sprint 6b follow-up)

Cross-engine / cross-restart scenarios (two `FlowEngine` instances sharing a `JdbcFlowSnapshotStore` + Postgres Testcontainer, mid-saga kill, Kafka rebalance during in-flight choreography) remain follow-up work for a later Sprint 6b PR.


arkstack-dev:
## Exeris Kernel — PR Review
**Triage:** LOCAL · Kafka consumer-loop empty-subscription guard; fix is driver-internal with no SPI or Core impact.

**Verdict:** APPROVE

**Strong pattern:** `startWithoutRegistrationsThenRegisterAndPublish` landing in `AbstractKafkaEventEngineTck` rather than the IT subclass — the race is now a contract obligation for every Kafka binding, not just the current one.

---
*False positive? Reply `@claude false-positive <id>` or thumbs-down.*

---

PR #94, opened 2026-05-08
v0.7 Sprint 6 telemetry: distributed-saga JFR events (ADR-013 §8)

arkstack-dev:
## Summary

ADR-013 §8 names three required JFR signals for distributed saga operations: wake-on-load fallback, optimistic-lock conflicts, and parked-enumeration latency. This PR delivers the first two; the third is deferred until a Core caller of \`listParked\` exists (today only the JDBC store implements it).

### \`WakeOnLoadFallbackEvent\` — \`eu.exeris.kernel.flow.WakeOnLoadFallback\` (Core)
Emitted by \`CoreFlowRuntime.lookupParked\` whenever the in-memory \`parkedInstances\` / \`liveInstances\` indices miss and the path falls back to the durable snapshot store. Operators read the rate to monitor cross-engine wake patterns; \`restored\` distinguishes a genuine cross-engine restore from a stale wake for an unknown instance, and \`loadDurationNanos\` surfaces slow durable-store reads.

Fields: \`engineName\`, \`instanceIdMost\`, \`instanceIdLeast\`, \`restored\`, \`loadDurationNanos\`.

### \`OptimisticLockConflictEvent\` — \`eu.exeris.kernel.flow.OptimisticLockConflict\` (Community/JDBC)
Emitted by \`JdbcFlowSnapshotStore.save\` on both race-loser paths from ADR-013 §5: \`phase=UPDATE_STALE\` (stale \`schemaVersion\` on UPDATE) and \`phase=INSERT_TOCTOU\` (first-writer INSERT race remapped from integrity-constraint violation). Carries the \`loadedSchemaVersion\` so operators can correlate conflicts with concrete saga rows.

Fields: \`engineName\`, \`phase\`, \`loadedSchemaVersion\`.

Both events follow the kernel JFR conventions: \`@StackTrace(false)\`, \`isEnabled()\` guard before commit, secret-safe payload (no opaque-state bytes, no JDBC connection strings, no Kafka credentials).

## Tests

- \`CoreFlowRuntimeTest\` — two new JFR-emission tests using \`RecordingStream\`: one asserts \`restored=true\` on the cross-restart fallback hit, one asserts \`restored=false\` on a wake for an unknown instance.
- \`CommunityJdbcFlowSnapshotStoreTckIT\` — two new integration tests that pin OCC JFR emission for both race-loser flavors against the real Postgres Testcontainer.

## Doc sync

- \`docs/subsystems/flow.md\` — JFR Events table extended with the two new events.
- \`docs/adr/ADR-013 Distributed Saga State Distribution Model.md\` §8 — marks both events as implemented and explains the deferred parked-enumeration JFR.

## Test plan

- [x] 0 PMD / 0 Checkstyle on touched modules.
- [x] Full unit suite green (only additions; existing tests unaffected).
- [x] All ITs green: 10 Kafka + 12 JDBC (10 inherited TCK tests + 2 new OCC-emission tests).

## Out of scope

- \`ParkedEnumerationLatencyEvent\` — deferred until a Core caller of \`listParked\` is wired (the contract is defined in SPI but not yet invoked from the engine).


arkstack-dev:
## Exeris Kernel — PR Review
**Triage:** LOCAL · JFR telemetry additions for two ADR-013 §8 signals; boundaries and SPI untouched.
**Verdict:** REQUEST CHANGES

### Findings

**[CONTRACT]** `CommunityJdbcFlowSnapshotStoreTckIT.java:emitsOccEventOnInsertToctou` — The test name and display name advertise INSERT_TOCTOU coverage, but the scenario is sequentially deterministic: `winnerStore.save(first)` commits before `loserStore.save(duplicate)` begins, so `tryOptimisticUpdate` finds 0 affected rows (version already bumped by the winner), then `existsInTransaction` sees the committed row, and the UPDATE_STALE branch is taken every time. The INSERT_TOCTOU branch inside `insertOrRemapPkConflict` — and its `OptimisticLockConflictEvent.emit(…, PHASE_INSERT_TOCTOU, …)` call — is never reached. The loose assertion (`.isIn("UPDATE_STALE", "INSERT_TOCTOU")`) acknowledges the uncertainty but doesn't fix it; this test is functionally a second copy of `emitsOccEventOnUpdateStale` with a different engine name.

**Fix options (pick one):**
1. Rename the test to reflect what it actually covers (`emitsOccEventOnDuplicateSave_takenAsUpdateStale`) and update the assertion to `.isEqualTo("UPDATE_STALE")`.
2. Keep the INSERT_TOCTOU intent and add a concurrent variant: two threads that each call `save` on a _brand-new_ instance simultaneously (no pre-existing row, coordinated so both pass the UPDATE check before either commits), then assert the loser emits `PHASE_INSERT_TOCTOU`. Note a `CountDownLatch` barrier between the `tryOptimisticUpdate` exit and the `insertOrRemapPkConflict` call would be needed to make the race deterministic.

Option 1 is the pragmatic fix for now; option 2 can be a follow-up TCK expansion.

---

**Strong patterns**

`@StackTrace(false)` on both events, `isEnabled()` guard inside each `emit()`, dual-mode `restored` flag in `WakeOnLoadFallbackEvent` cleanly separating cross-engine-restore from stale-wake — all exactly right. Timing wraps the actual store call so `loadDurationNanos` reflects I/O cost, not event overhead.

---

*False positive? Reply @claude false-positive or thumbs-down.*

---

arkstack-dev:
Confirmed the reviewer's analysis — `emitsOccEventOnInsertToctou` deterministically took UPDATE_STALE, never INSERT_TOCTOU, because the winner's commit was always visible to the loser's `existsInTransaction` under READ COMMITTED.

Applied **Option 1** (pragmatic rename + tighten) in `f3b7579`:

- Method renamed: `emitsOccEventOnInsertToctou` → `emitsOccEventOnDuplicateSaveTakenAsUpdateStale`
- `@DisplayName` updated to match.
- Assertion tightened: `.isIn("UPDATE_STALE", "INSERT_TOCTOU")` → `.isEqualTo("UPDATE_STALE")`.
- Inline comment corrected — replaced the inaccurate "depends on transaction scheduling" with an explanation of why INSERT_TOCTOU is not reachable from this sequential setup, plus a pointer to PR #94's review for the deferred genuine-race coverage.

The renamed test still adds non-redundant coverage alongside `emitsOccEventOnUpdateStale`: it verifies the JFR `engineName` field surfaces the loser's store name, not the winner's, which matters for cross-engine deployments where multiple engines share the durable store.

Genuine INSERT_TOCTOU JFR coverage (concurrent first-writer race with a `CountDownLatch` barrier between `tryOptimisticUpdate` and `insertOrRemapPkConflict`) is left as a follow-up TCK extension as the reviewer suggested.

Both renamed tests pass against the Postgres Testcontainer; 0 PMD / 0 Checkstyle on the touched module.

arkstack-dev:
## Exeris Kernel — PR Review
**Triage:** LOCAL · JFR telemetry additions for two ADR-013 §8 signals; boundaries and SPI untouched.
**Verdict:** APPROVE

Previous [CONTRACT] finding addressed in `f3b7579`: test renamed to `emitsOccEventOnDuplicateSaveTakenAsUpdateStale`, assertion tightened to `.isEqualTo("UPDATE_STALE")`, inline comment corrected to accurately describe the sequential-determinism reason INSERT_TOCTOU is unreachable from this setup.

**Strong patterns**

`@StackTrace(false)` on both events; `isEnabled()` guard before field writes in both `emit()` methods; `loadDurationNanos` carries real I/O latency via a `System.nanoTime()` bracket outside `emit()`, not event-construction overhead; `restored` dual-mode flag emitted unconditionally so operators can distinguish genuine cross-engine restores from stale wakes in a single event stream query; Wall placement correct (`WakeOnLoadFallbackEvent` in Core, `OptimisticLockConflictEvent` in Community/JDBC).

---

PR #148, opened 2026-05-29
Fix ConnectionAcquireEvent handling and improve admission control

arkstack-dev:
## Summary

Two persistence runtime-safety fixes for the **0.7.1** patch line, both surfaced by running
the acquire/admission paths under realistic Loom + CPU-constrained conditions. Neither touches
the SPI/Core boundary in a breaking way (the only SPI change is Javadoc).

| Commit | Fix |
  |---|---|
| `19117ad1` | JFR `ConnectionAcquireEvent` made single-phase — eliminates a JDK 26 GA SIGSEGV when the acquire path runs on a virtual thread |
| `4b1897a9` | Persistence admission control made tunable + recalibrated for small pools (ADR-035) — fixes the ~92% 503 regression on the constrained benchmark |

15 files, +956 / −274.

  ---

### 1. `ConnectionAcquireEvent` VT JFR crash (`19117ad1`)

`ConnectionAcquireEvent` was `begin()`'d before the blocking pool checkout in
`CommunityHikariSupport.acquireConnection` and `commit()`'d after it. On a virtual thread the
checkout parks and unmounts the carrier; remounting on a **different** carrier and committing the
held event flushes a stale, carrier-bound `JfrBuffer` → **SIGSEGV in `JfrStorage::flush_regular_buffer`**
(reproduced deterministically on JDK 26 GA `build 26+35`, `settings=profile`, under pool contention;
JFR-off runs were unaffected).

Not a JDK flake and not a persistence regression — the event has existed since v0.5.0; it is the
**held-event straddle** exposed by running the acquire path on virtual threads.

**Fix:** the event is now constructed, populated, and committed entirely *after* the checkout returns
(single-phase, post-mount carrier, never held across the unmount). Acquire latency moves into an
explicit `acquireLatencyNs` field, mirroring the safe discrete-commit pattern of `TransactionLifecycleEvent`.

**Coverage:** `CommunityConnectionAcquireJfrVirtualThreadTest` — parked acquire on a VT + 32-VT acquire
churn under an active JFR recording; asserts one event per acquire with correct fields and no crash
across the unmount/remount boundary.

  ---

### 2. Tunable admission control + small-pool recalibration — ADR-035 (`4b1897a9`)

The constrained 1-vCPU `entity-read-by-id` benchmark regressed to **~92% 503s since v0.6.0**. Root cause:
`canServiceRequest()` (the 503 gate added in v0.6.0) shed on the *first* queued acquire, and the adaptive
pool collapses to **2 connections** under `-XX:ActiveProcessorCount=1`. v0.5 (no gate) queued briefly and
returned **0 errors** at an even tighter envelope — so it is a behavior regression, not a resource limit.

- **`CommunityAdmissionConfig`** — tunable thresholds (`persistence.admission.*`) + new
  `queueDepthAllowanceRatio`; `@Dynamic` hot-reloadable `CURRENT` (the kernel's **first production
  `@Dynamic` consumer**: Community startup-configurable, Enterprise hot-reload).
- **Recalibrated rule** — a full/saturated pool sheds only when
  `pendingAcquires > ceil(maxPool × queueDepthAllowanceRatio)`. Default `8.0` → `maxPool=2` admits the
  16-client burst (restores 0% errors); `0.0` restores the strict pre-035 contract.
- **Contract change** — "shed on a forming queue" is demoted from a universal SPI contract to a tunable
  tier policy. `AbstractPersistenceEngineAdmissionControlTck` is relaxed to the cross-tier invariants
  (capacity → admit; closed → shed; non-blocking; consistent); strict shed assertions move to the
- **Doc-drift fix** — `PersistenceEngine#canServiceRequest` Javadoc cited "ADR-010", which is actually
  *Host Runtime Model* (exeris-spring-runtime); now cites ADR-035.

**Coverage:** 44 admission tests green (TCK + `CommunityAdmissionConfigTest` + Fairness/CanServiceRequest
+ `HttpDispatcherBackpressureTckTest`), including the small-pool-admits regression guard and the
  deep-queue-still-sheds backpressure guard.

  ---

## Test plan

- [x] `mvn -pl exeris-kernel-community -am test -Dtest=…Admission…,…CanServiceRequest…,…Fairness…,…Backpressure…` → 44 run, 0 failures, 0 errors
- [x] JFR-VT regression test passes under an active recording (no SIGSEGV across unmount/remount)
- [x] **Recommended before merge:** re-run the constrained `entity-read-by-id` benchmark (PRE ≈ 92% errors → POST ≈ 0%); confirm reject reasons disappear via `AdmissionDecisionEvent`/JFR

## Notes

- ADR-035 registry row to be landed in `exeris-docs/adr-index.md` (separate repo, separate commit).
- Out of scope (separate `exeris-benchmarks` follow-up): the harness reports `runner_status: success` +
  inflated `throughput_rps` despite a 92% error rate — high-error runs should be gated/flagged as invalid.


arkstack-dev (commented):
## Exeris Kernel — PR Review
**Triage:** LOCAL · two persistence runtime-safety fixes; SPI change is Javadoc-only; The Wall intact.
**Verdict:** APPROVE

---

### Findings

**[PERFORMANCE]** `CommunityPersistenceAdmissionController.java` — `config.queueDepthAllowance(max)` executes `(int) Math.ceil(maxConnections * queueDepthAllowanceRatio)` on **every** `canServiceRequest()` call. For a given config snapshot, `maxConnections` is stable (pool-configured max doesn't change at runtime) and `queueDepthAllowanceRatio` is immutable in the record, so the result is constant between config swaps. Pre-computing `allowance` into `CommunityAdmissionConfig` at construction time — or adding a cached `int queueDepthAllowance` field alongside the ratio — would reduce the hot-path admission decision to a single `int` compare instead of a FP multiply + `Math.ceil` per call. Low urgency but worth doing before the release-merge review.

**[OBSERVATION — test only]** `CommunityConnectionAcquireJfrVirtualThreadTest` line 994: `parked.countDown()` fires *before* the `acquireConnection()` call in the VT, so `Thread.sleep(200)` is a timing heuristic for "VT has reached the blocking checkout." If the VT hasn't parked within 200 ms the unmount/remount path isn't exercised. The test class Javadoc correctly acknowledges the structural guarantee (not the crash itself) is what's being locked in, so this isn't a defect — but a spin on `pool.getHikariPoolMXBean().getThreadsAwaitingConnection() > 0` would make the scenario deterministic. Non-blocking at this tier.

---

**Strong patterns**

`ConnectionAcquireEvent` migrated to discrete single-phase commit with `@StackTrace(false)` — eliminates the carrier-bound `JfrBuffer` stale-flush crash on virtual-thread unmount/remount; correctly mirrors the `TransactionLifecycleEvent` pattern. `@Dynamic static volatile CURRENT` with immutable-record swap is the right hot-reload idiom — zero-lock, no heap churn on the decision path.

---

*False positive? Reply `@claude false-positive <id>` or 👎*

---

arkstack-dev:
**Exeris Kernel — Automated Review**
**Verdict:** APPROVE (with two contract observations for v0.7.1 / v0.8 follow-up)
**Triage:** MULTI_DOMAIN release-cut · primary risk: admission-control contract loosening reaches further than the stated "small-pool admits transient queue" goal.
**Scope reviewed:** `docs/ROADMAP.md` (v0.7 closeout), `docs/adr/ADR-035-*`, `docs/subsystems/persistence.md`, persistence SPI (`PersistenceEngine`), persistence Core (`ConnectionAcquireEvent`), persistence Community (`CommunityAdmissionConfig`, `CommunityPersistenceAdmissionController`, `CommunityHikariSupport`, `CommunityPersistenceEngine`, `CommunityPersistenceSubsystem`), TCK (`AbstractPersistenceEngineAdmissionControlTck`), and the four touched Community tests. Pre-existing PR comments: only the GH-Actions dependency review and SonarCloud bot (quality gate passed) — no prior Exeris-Kernel comment for head `d558c6f` to update.

### Findings

- [CONTRACT] `exeris-kernel-community/.../CommunityPersistenceAdmissionController.java:55-75` — `evaluateAdmissionReason` now gates **both** the `REJECT_HARD_SATURATION` branch and the entire `REJECT_GUARD_BAND_FAIRNESS` branch behind `queueExceedsAllowance` (`queued > ceil(max × ratio)`, default 8.0). Effect under DEFAULT: a pool with `max=10` only shed-fires once `queued > 80`; `max=2` needs `queued > 16`. Until then, `FairnessTracker.indicatesAdmissionStress(...)` and `shouldRejectEarlyInGuardBand(...)` are unreachable. ADR-035 frames this as restoring small-pool availability, but it silently neutralizes the fairness-inversion machinery for typical pool sizes, not just constrained 1-vCPU profiles. Fix: either keep the fairness/guard-band path queue-independent (or gated by a *separate*, smaller threshold) so sustained fairness stress can still trigger before the queue is ~8× the pool, or document this as an intentional retirement of `FairnessTracker` under the recalibrated default in ADR-035 §Consequences. — ref: ADR-035 ("recalibrated rule … shed only once `pendingAcquires > queueDepthAllowance(maxPool)`") and `docs/subsystems/persistence.md` §"Admission Control & Backpressure Integration".

- [TCK DEBT] `exeris-kernel-tck/.../AbstractPersistenceEngineAdmissionControlTck.java:120-160, 195-225, 385-405` — TEST 2 / TEST 3 / `handlesSmallPoolSize` were relaxed to `assertThat(first).isEqualTo(second)`. The universal TCK now exercises **no** shed direction except `engine.close()`. An Enterprise binding that always returns `true` is conforming. The deterministic reject-reason machine moved entirely to Community-only tests via `CommunityAdmissionConfig.STRICT` — there is no out-of-repo Enterprise binding obligation recorded for it. Fix options: (a) add an opt-in TCK hook (e.g. `@Tag("strict-admission")` subclass override) that lets bindings declare "I do shed on deep queue" and assert that path; or (b) record an explicit Enterprise binding obligation in ADR-035 §Consequences and reference it from the TCK Javadoc. — ref: CLAUDE.md §"Review Priorities #4 (Verification impact proportional to behavior change)", `docs/modules/05-tck.md`.

- [HEURISTIC] `exeris-kernel-community/src/test/java/.../HttpDispatcherBackpressureTckTest.java:84-118` mutates the process-global `CommunityAdmissionConfig.CURRENT` with save/restore in finally. Correct today (Surefire forks per class) but fragile if test parallelism is ever enabled. Same pattern recurs in `CommunityPersistenceEngineFairnessTest` (which side-stepped this by constructing the controller directly with `STRICT`); consider promoting that pattern, or routing the test config through `CommunityPersistenceEngine.canServiceRequest(snapshot)` with a per-instance config holder.

### Notes

Strong patterns to keep: (1) `ConnectionAcquireEvent` single-phase commit is the right structural fix — the analysis of carrier-bound `JfrBuffer` flush across VT unmount/remount is precise, the regression test exercises the actual triggering condition (parked acquire on a VT under an active `Recording`), and the SPI Javadoc honestly states the contract change `MUST → SHOULD` rather than hand-waving. (2) ADR-035 keeps the entire tunable surface in `CommunityAdmissionConfig` with no `PersistenceConfig` (SPI) field added — The Wall is preserved (verified: no `CommunityAdmissionConfig` references in `exeris-kernel-spi` or `exeris-kernel-core`). (3) `STRICT` as a named, validated record literal that exactly reproduces pre-035 semantics gives operators a deterministic rollback path and gives tests a deterministic reject-reason machine. (4) ADR-010 → ADR-035 Javadoc reference is correctly redirected. ROADMAP carries no v0.7.1 section; v0.7 closeout entries remain accurate — no drift introduced.

---
*Generated automatically. False positive? Reply `@claude false-positive <id>` or thumbs-down the comment.*

---

PR #156, opened 2026-06-03
v0.8 Sprint 7 (GRAPH-111): bind ExecutionGraphZeroAlloc + ChurnRatio TCKs

arkstack-dev:
## Sprint 7 · GRAPH-111 — Graph baseline production hardening

Closes the documented Graph TCK gap (`graph.md`): the two Graph contract TCKs had **zero Community-tier bindings**, so neither was gated in CI. Scope-disciplined per ROADMAP ("avoid broadening beyond practical paths already in use") — **no SPI/Core/`maven.yml` changes**, only bindings + the abstract-TCK bug fix they surfaced.

### Bindings
- **`CommunityExecutionGraphZeroAllocTckTest`** — unit, no DB. The single-arg `GraphSession.findShortestPath(src,tgt)` is fully in-process (distinct ids → `PathResult.notFound`), so the shortest-path hot path is measured in the main `build-and-verify` lane.
  - **Measured: 0 `eu.exeris.*` allocs/iter** (budget 10) over 1000 iters.
- **`CommunityGraphChurnRatioTckIT`** — `@Tag("integration")` + Neo4j Testcontainers, Cypher backend (built from an explicit client via `GraphConfig`, so no `ScopedValue`-bound persistence engine is needed to drive the traversal). Runs automatically in the existing `persistence-rls-gate`.
  - **Measured churn-to-data ratio ~10.2x** against the documented `<20x` Community SLO (baseline ~15x).

### Latent abstract-TCK bug fixed (surfaced by the binding)
`GraphChurnRatioTck` read `allocationSize` unconditionally, but on JDK 26 the recording is dominated by `jdk.ObjectAllocationSample` events, which carry no `allocationSize` (they expose `weight`) — the unconditional `getLong` threw `IllegalArgumentException` before any ratio could be computed. Never caught because the TCK had no concrete binding (exactly the gap GRAPH-111 closes). Fix reads `allocationSize` when present (TLAB), else `weight` (sampler), else fails fast on a genuine schema change. Generic to every binding (incl. Enterprise), so the abstract TCK is the correct home. No threshold/iteration/expected-bytes change.

### CI lanes (verified)
- ZeroAlloc → `build-and-verify` (default lane).
- ChurnRatio → `persistence-rls-gate` (runs all `exeris-kernel-community` `@Tag("integration")` tests).

Exit criterion "Graph baseline TCK passes in CI" met. Remaining Sprint 7 item: FLOW-110.

### DIFF
--- a/docs/subsystems/graph.md
+++ b/docs/subsystems/graph.md
@@ -212,7 +212,7 @@ connections of user X"). Leaving it at the default 10 prevents runaway traversal
 - **Carrier Pinning:** JFR-based validation that driver I/O does not stall Virtual Threads
   (`CarrierPinnedEvent` must not fire during standard traversal).
 
-> **TCK gap:** `GraphChurnRatioTck` and `ExecutionGraphZeroAllocTck` have no Community-tier concrete bindings in `exeris-kernel-community/src/test/`. Community churn ratio enforcement (`EX-GRPH-5005`) is currently not gated in CI for the Community implementation.
+> **TCK bindings:** Both `ExecutionGraphZeroAllocTck` and `GraphChurnRatioTck` now have Community-tier concrete bindings in `exeris-kernel-community/src/test/`. `CommunityExecutionGraphZeroAllocTckTest` runs in the main `build-and-verify` lane (in-process shortest-path hot path, no database). `CommunityGraphChurnRatioTckIT` is `@Tag("integration")` (live Neo4j via Testcontainers) and runs in the `persistence-rls-gate` CI job — gating Community churn-to-data ratio (`EX-GRPH-5005`) below the documented 20x threshold.
 
 ---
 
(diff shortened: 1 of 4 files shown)


arkstack-dev (commented):
## Exeris Kernel — PR Review
**Triage:** LOCAL · Test-only: two Community TCK bindings + abstract TCK JFR bug fix; no production runtime code touched.

**Verdict: APPROVE**

**Strong pattern:** `GraphChurnRatioTck.java` uses `hasField()` guards + a fail-fast `AssertionError` on unknown event types rather than a try-catch — preserves the silent-false-negative protection against genuine JFR schema drift while correctly handling the `jdk.ObjectAllocationSample` / `weight` vs TLAB / `allocationSize` split on JDK 26.

---
*False positive? Reply `@claude false-positive` or thumbs-down.*

---

PR #186, opened 2026-06-15
feat(config): @Immutable sealed-key annotation + compile-time guard + watcher refusal (v0.9 Sprint 5)

arkstack-dev:
## Summary

Realizes the `@Immutable` config-key contract already documented in `docs/subsystems/config.md` — the deliberate inverse of `@Dynamic`. It seals a `static final` config key (security trust anchors, tenant isolation boundaries, native library paths) against any runtime hot-reload, enforced both at compile time and at runtime. Additive; **no new ADR** (it's the realization of a contract `config.md` already promised for v0.9 Sprint 5).

Closes the v0.9 ROADMAP entry *"Config: `@Immutable` Annotation Enforcement for Config Keys"*.

## What landed

**SPI** (`exeris-kernel-spi`)
- `Immutable` — `@Target(FIELD)`, RUNTIME, `file`/`key`/`reason` members.
- `ConfigProvider.guardImmutable(file, key)` — symmetric `default` no-op beside `watch()` (Community runs no watcher → already sealed; `RegistryBackedConfigProvider` records the guard).
- `KernelErrorCodes.EX_CFG_1004` — secret-safe refusal code (file + key, **never** the value).

**Core** (`exeris-kernel-core`)
- `KernelConfigRegistry` — sealed-key guards (`registerImmutable` / `isImmutable` / `immutableRegistrations`) under the same `VarHandle`-sealed discipline as `@Dynamic`.
- `DynamicConfigFileWatcher` — captures each sealed key's boot value in `start()` **before** the watcher VT spawns (eliminates a seed-vs-event race); on a later on-disk change it **refuses** the reload, preserves the sealed value, and emits `ImmutableReloadRefusedEvent` (`EX-CFG-1004`). Defense in depth: a contradictory `@Dynamic` binding that ever reached runtime loses to the sealed intent.
- `ImmutableReloadEvent` — single-phase JFR commit (file + key only).

**build-config** (`exeris-kernel-build-config`)
- `ImmutableConfigProcessor` — validation-only APT processor (reads annotation FQNs, no SPI reactor edge — mirrors `RequiresRoleProcessor`). Fails the build on the mutually-exclusive `@Immutable`+`@Dynamic` combo and on a non-`static final` placement. Registered via `META-INF/services`.

**Docs** — `config.md` (realization + `EX-CFG-1004` + security-relevant key catalog for GA), `ROADMAP.md` (status DELIVERED), `stability-matrix.md` (`preview` until the `AbstractConfigProviderTck` binding lands).

## Merge gate (from ROADMAP)
- ✅ APT processor catches the mutually-exclusive annotation combo — `ImmutableConfigProcessorTest` (3 cases).
- ✅ Runtime watcher refuses `@Immutable` re-read with a structured event — `DynamicConfigFileWatcherTest.ImmutableKeyRefusal` (2 cases, asserts `EX-CFG-1004` JFR).
- ✅ Documented `@Immutable` key catalog — `config.md` §"Security-Relevant Key Catalog".

## Tests
- `ImmutableConfigProcessorTest` — 3/3 (both-annotations / non-final / valid).
- `DynamicConfigFileWatcherTest.ImmutableKeyRefusal` — 2/2 (sealed refused while sibling `@Dynamic` reloads; JFR asserted).
- `KernelConfigRegistryTest.ImmutableGuardContract` — 4/4 (register / null-file match / ignored-after-seal / unmodifiable snapshot).
- No regressions in config/bootstrap suites. Lint: **0 checkstyle / PMD pass** on spi+core. Downstream consumers test-compile clean.

## Notes
- **The Wall held** — SPI uses only `java.lang.annotation`; `guardImmutable` is a `default` so no existing `ConfigProvider` / test-double needs changes.
- `TooManyMethods` on `DynamicConfigFileWatcher` suppressed with justification (two cohesive enforcement domains on one watch loop — `@Dynamic` reload + `@Immutable` refusal — not inflation).
- Follow-up: a dedicated `AbstractConfigProviderTck` binding to promote `@Immutable` from `preview` → `stable` (separate PR).


arkstack-dev:
Addressed the required fix (`2251c312`) — with a correction to the review's premise:

**The separator was already `'\0'` (NUL), not a space.** `baselineKey()` returned `fileName + '\0' + key`, matching the `immutableBaselines` Javadoc, so the pathological filename collision the review describes (`"a b"/"c"` vs `"a"/"b c"`) could never actually occur — NUL cannot appear in a POSIX filename or a dot-notation key.

**The real issue:** the NUL char literal was encoded as a **raw 0x00 byte** in the source rather than the `'\0'` escape sequence. That made the entire file register as *binary* to `grep`/`diff`/review tooling — which is almost certainly why the review read the separator as a space (the tool couldn't display the NUL). One raw NUL byte confirmed via hex scan; replaced with the readable `'\0'` escape.

Result: **byte-identical runtime behavior** (still NUL-separated, still collision-proof) + clean ASCII source that tools can read. Config watcher + registry tests (30, incl. `ImmutableKeyRefusal`) green; pmd/checkstyle clean.

Non-blocking items left as follow-ups per the review:
- Advisory: `refuseImmutableReloads` does N `extract()` reads for N sealed keys in one file — negligible on the watcher (non-hot) path; single-pass extraction is a clean follow-up.
- TCK gap: `AbstractConfigProviderTck` binding for `guardImmutable()` — correctly deferred; `@Immutable` stays marked preview in the stability matrix until then.

Want me to also pin the NUL choice with a tiny regression test (assert `baselineKey` is collision-safe for space-containing filenames), to stop a future "cleanup" from reintroducing a space separator?

arkstack-dev:
Added the collision regression test (`53a24be6`) — with one deliberate deviation from the suggested form.

**Chose the pure-function injectivity unit test, not the integration route.** Reason: the colliding sibling key in the scenario (`"anchors.properties a"`) **cannot be constructed via a real `.properties` file** — whitespace terminates a key in `java.util.Properties` parsing, so `anchors.properties a=v` parses as key `anchors.properties`. A single-registration integration test (as sketched) never actually creates the collision, so it would pass under **both** `'\0'` and `' '` — pinning nothing. The unit assertion exercises the exact injectivity property and genuinely fails under a space separator.

To enable it without reflection or widening public API, `baselineKey` is now package-private (`/* default */ static`) — the test is in the same package (`eu.exeris.kernel.core.config`). Added a Javadoc note tying the separator choice to the guard test.

```java
assertThat(DynamicConfigFileWatcher.baselineKey("security anchors.properties", "a"))
    .isNotEqualTo(DynamicConfigFileWatcher.baselineKey("security", "anchors.properties a"));
```

Watcher tests **13/13** (ImmutableKeyRefusal 2→3), pmd/checkstyle clean. Both non-blocking follow-ups (single-pass extraction, AbstractConfigProviderTck) remain out of scope as agreed.

PR #187, opened 2026-06-15
fix(crypto): terminate TLS handshake on fatal SSL_get_error to stop reactor busy-spin

arkstack-dev:
## Problem

`OffHeapTlsEngine.mapSslError` mapped every non-WANT `SSL_get_error` code — including the **fatal** `SSL_ERROR_SSL` (protocol error, e.g. non-TLS bytes on the TLS port) and `SSL_ERROR_SYSCALL` (I/O error / unexpected peer EOF) — to `NEED_HANDSHAKE` (retry).

On a non-blocking fd-owner BIO, a "would block" condition always surfaces as `WANT_READ`/`WANT_WRITE`, so any other code is terminal. A half-open or garbage probe connection stuck in `HANDSHAKE_IN_PROGRESS` was therefore re-stepped on **every level-triggered read**, never torn down — busy-spinning a reactor thread until the 10s handshake timeout and emitting **~1.4M phantom `HandshakeFailure` JFR events** per run.

Root-caused from a JFR capture: `1,435,151` `HandshakeFailure` events from only **2 distinct `sslPtr`** (2 connections), `ConcurrentLinkedQueue` poll/offer at **27% of CPU**.

## Fix

- `mapSslError`: `default -> CLOSED` (fatal/unknown codes are terminal). `ensureTlsReady()`/`unwrap()` already tear the fd down on `CLOSED`, so a bad connection now closes after **one** real failure instead of spinning.
- Add `SSL_ERROR_SSL` / `SSL_ERROR_SYSCALL` constants to `CoreOpenSslLoader`.
- Make `mapSslError` package-private `static` + add a pure mapping unit test (no OpenSSL needed).
- Align the **Core** crypto TCK with the existing **Community** TCK: the fd-owner engine must fail-fast before external bind. The previous artificial `notifyBound()`-without-fd setup masked the bug through the "beginHandshake never immediately returns CLOSED" invariant.

## Validation (wrk + TLS loopback, same driver/config, before → after)

| metric | before | after |
|---|---|---|
| throughput | 3493 rps | **4755 rps (+36%)** |
| cpu / req | 1.29 ms | **0.43 ms** (Quarkus parity) |
| RSS avg | 1327 MB | **594 MB** |
| p99 latency | 73.6 ms | **54.7 ms** |
| JFR `HandshakeFailure` | 1,435,151 | **3** |

In the post-fix JFR the `ConcurrentLinkedQueue` poll/offer pair drops out of the top hot methods entirely; the hot path is now real work (Jackson serialization + Postgres JDBC).

## Tests

- New `OffHeapTlsEngineSslErrorMappingTest` — 6/6.
- `CoreOffHeapTlsEngineTckTest` — 15 run / 0 fail (3 IO-gated skips).
- `CommunityKernelCryptoProviderTckTest` — 15 run / 0 fail (unchanged).
- PMD + Checkstyle (core) clean.

## Out of scope (separate follow-ups)

- Concurrent-handshake serialization on the single acceptor thread (connection-establishment latency, not steady-state throughput).
- Residual app-layer throughput gap vs Quarkus (Jackson/JDBC).

### DIFF
--- a/exeris-kernel-core/src/main/java/eu/exeris/kernel/core/crypto/openssl/CoreOpenSslLoader.java
+++ b/exeris-kernel-core/src/main/java/eu/exeris/kernel/core/crypto/openssl/CoreOpenSslLoader.java
@@ -50,11 +50,19 @@ import static java.lang.foreign.ValueLayout.JAVA_LONG;
 @SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.TooManyMethods"})
 public final class CoreOpenSslLoader {
 
-    /** {@code SSL_FILETYPE_PEM = 1} from {@code openssl/ssl.h}. */
+    /**
+     * {@code SSL_FILETYPE_PEM = 1} from {@code openssl/ssl.h}.
+     * Distinct OpenSSL domain from {@link #SSL_ERROR_SSL} despite the shared value {@code 1}:
+     * this is a file-type argument, not an {@code SSL_get_error} code.
+     */
     public static final int SSL_FILETYPE_PEM   = 1;
     /** {@code SSL_VERIFY_NONE = 0} — no peer certificate verification. */
     public static final int SSL_VERIFY_NONE    = 0;
-    /** {@code SSL_ERROR_SSL = 1} — fatal protocol error (e.g. non-TLS bytes on a TLS port). */
+    /**
+     * {@code SSL_ERROR_SSL = 1} — fatal protocol error (e.g. non-TLS bytes on a TLS port).
+     * An {@code SSL_get_error} return code; unrelated to {@link #SSL_FILETYPE_PEM} (same value {@code 1},
+     * different OpenSSL domain).
+     */
     public static final int SSL_ERROR_SSL        = 1;
     /** {@code SSL_ERROR_WANT_READ = 2}. */
     public static final int SSL_ERROR_WANT_READ  = 2;
--- a/exeris-kernel-core/src/main/java/eu/exeris/kernel/core/crypto/openssl/CoreOpenSslLoader.java
+++ b/exeris-kernel-core/src/main/java/eu/exeris/kernel/core/crypto/openssl/CoreOpenSslLoader.java
@@ -54,10 +54,14 @@ public final class CoreOpenSslLoader {
     public static final int SSL_FILETYPE_PEM   = 1;
     /** {@code SSL_VERIFY_NONE = 0} — no peer certificate verification. */
     public static final int SSL_VERIFY_NONE    = 0;
+    /** {@code SSL_ERROR_SSL = 1} — fatal protocol error (e.g. non-TLS bytes on a TLS port). */
+    public static final int SSL_ERROR_SSL        = 1;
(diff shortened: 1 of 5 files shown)


arkstack-dev:
Addressed the review in commit `adb72966`:

**#7 state self-guard (strongly recommended) — done.** `beginHandshake()` and `unwrap()` now call the existing `stateMachine.forceError()` seam when the result maps to `CLOSED`. A re-entrant `beginHandshake()` on a broken session then fails fast on the wrong-state guard (`phase != HANDSHAKE_IN_PROGRESS`) instead of re-driving `SSL_accept`, so teardown no longer depends solely on the transport honouring the `CLOSED` status. `WANT_READ`/`WANT_WRITE` stay non-terminal (no state change, hot path unaffected — one enum comparison). Applied to `unwrap()` too for the consistency the review noted.

**#8 nit — done.** Cross-reference Javadoc added on `SSL_FILETYPE_PEM` and `SSL_ERROR_SSL` (shared value `1`, unrelated OpenSSL domains).

**#6 IT coverage — confirmed gap, NOT closed by this PR.** `OffHeapTlsEngineLoopbackIT` covers the happy path only (handshake → ACTIVE, 512-byte round-trip, SESSION-slab backpressure). It does **not** exercise the garbage-probe / non-TLS-bytes → `SSL_ERROR_SSL` → `CLOSED` → terminal lifecycle. The new `OffHeapTlsEngineSslErrorMappingTest` locks the mapping table, but the full fatal-path teardown is not integration-tested. Recommend a fast-follow IT: plain-TCP client sends non-TLS bytes to the server engine, assert `beginHandshake()` returns `CLOSED` and the engine reaches `ERROR`.

Re-verified after the change: crypto + state-machine + loopback IT suites green (`CoreOffHeapTlsEngineTckTest` 15/0fail, `TlsStateMachineTest` all green, `OffHeapTlsEngineLoopbackIT` 3 run, `OffHeapTlsEngineSslErrorMappingTest` 6/6); PMD + Checkstyle (core) clean.

arkstack-dev:
**#6 (IT coverage gap) — now closed** in `ea2b6a79`.

`OffHeapTlsEngineLoopbackIT.garbageProbeYieldsClosedAndErrorState` drives a real loopback handshake where a plain-TCP client sends non-TLS bytes (HTTP probe) to the TLS port, and asserts:
- `beginHandshake()` returns terminal `CLOSED` (was the retryable `NEED_HANDSHAKE` that busy-spun),
- the engine self-guards to `TlsPhase.ERROR`,
- `isHandshakeComplete()` is `false`,
- a re-entrant `beginHandshake()` fails fast with `TlsHandshakeException` instead of re-driving `SSL_accept`.

Verified under Failsafe (the IT's real lifecycle, which supplies `--add-opens java.base/sun.nio.ch=ALL-UNNAMED`): **4/4 green, 0 skipped** — the existing handshake/round-trip/session-slab ITs plus the new fatal-path test. PMD + Checkstyle (core) clean.

All three review items (#6, #7, #8) are now addressed in-branch.

PR #188, opened 2026-06-15
fix(transport): deserialize TLS handshakes — reactor-driven established hook (fix #2)

arkstack-dev:
## Problem

The single platform acceptor thread (`NativeTcpCarrier.runAcceptorLoop`) blocked on **each connection's full TLS handshake** — `registerAndHandshakeConnection` → `awaitHandshakeReadyForConnection()` → `NativeTcpStream.ensureTlsReady(true)` (a `parkNanos(250µs)` loop) — **before** `accept()`ing the next. All TLS handshakes serialized on one thread.

Measured (h2load, 128 conns, validated): connect-time **mean 503ms / max 1.29s** vs Quarkus **63ms** (8x). JFR: median inter-accept gap **6.0ms** == single-handshake wall time; individual handshakes are fast (0–6.5ms). The blowup is pure serialization, hitting **connection-establishment latency** (cold-start / reconnect-storm / churn), not steady-state throughput.

## Fix (reactor-driven, architect-recommended Option B)

The reactor already drives the handshake non-blockingly (`readTlsIngressFromFd` → `ensureTlsReady(false)`). This removes the redundant acceptor-side blocking driver and fires a **one-shot established hook** from the reactor instead:

- **Plaintext** — fired when the reactor arms the key (`markRegistrationReady`).
- **TLS** — fired when the reactor-driven handshake reaches ACTIVE (`readTlsIngressFromFd`, after the first `ensureTlsReady(false)==true`).

The hook runs `onConnectionEstablished` (contractually non-blocking, carrier-thread per the `ConnectionHandler` SPI) then `paqs.schedule`, preserving order. A one-shot `AtomicBoolean` CAS latch guarantees exactly-once. The acceptor returns to `accept()` immediately → all 128 handshakes progress concurrently on the reactor pool.

## Boundary / architecture

- **Community-internal.** No SPI/Core change, no ADR (refactor that *converges on* the TCK-064 reactor-driven direction — "no separate ingress/writer VT survives").
- **No second STS exception** — unlike a per-connection VT spawn, this adds no unstructured-concurrency site (Option A was rejected for that reason).
- Slot accounting stays lifecycle-based (released on stream close); a failed/aborted handshake closes the stream and releases the slot without acceptor involvement.
- Removes the now-unused `awaitRegistrationReadyForConnection` / `awaitHandshakeReadyForConnection`.

## Tests

- New `connectionEstablishedFiresExactlyOnceAndBeforeStreamDispatch` — asserts the hook fires **exactly once** under multiple ingress reads and **before** the first stream dispatch.
- Transport suite: **73 run / 0 fail** (TCK contract bindings + multi-reactor + stress).
- E2E client/server + ingress integration tests green via Failsafe (TLS + plaintext establishment paths).
- PMD + Checkstyle (community) clean.

## Acceptance signal (follow-up)

Re-run the h2load 128-conn connect-time benchmark in `exeris-benchmarks` — expectation is connect mean collapsing from ~503ms toward the single-handshake floor (~6ms). Benchmarks stay non-gating per ecosystem rule.

### DIFF
--- a/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/NativeTcpCarrier.java
+++ b/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/NativeTcpCarrier.java
@@ -474,7 +474,7 @@ public final class NativeTcpCarrier implements TransportEngine {
                 NativeTcpStream stream = buildAcceptedStream(currentChannel, connection);
                 connection.bindSingleStream(stream);
                 connectionManagedByStreamLifecycle = true;
-                registerAndHandshakeConnection(connection, stream, currentChannel);
+                registerConnection(connection, stream, currentChannel);
             } catch (RuntimeException exception) {
                 if (connection != null) {
                     connection.close();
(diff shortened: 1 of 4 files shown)


arkstack-dev:
Addressed the review in `04c712c0` (all findings were non-blocking):

**#3 + #1 — JFR observability for the established hook.** New `ConnectionEstablishedEvent` fired once per connection from `fireEstablishedOnce()`:
- `establishLatencyNanos` — key-registration → established (≈0 plaintext, handshake duration for TLS): makes the connect-latency win visible in production recordings.
- `handlerDurationNanos` — time the handler + schedule spent on the reactor thread. This directly surfaces the **#1** concern: a `ConnectionHandler` that blocks the selector loop now shows up as a large `handlerDurationNanos`, rather than relying on convention alone.

Single-phase `commit()` on the reactor platform thread (no `begin→blocking→commit` straddle), `FlightRecorder.isInitialized()` guard for zero overhead when off, fired once per connection (off the request hot path).

**#2 — `completeEstablished()` now logs** the previously-swallowed `onConnectionEstablished` and `paqs.schedule` failures at WARNING before closing.

**#4 — establishment test** now polls for one-shot stability (count never exceeds 1, settles at 1) instead of the fixed 100 ms park.

**#5 — TLS-isolated one-shot test: deferred (documented).** The latch is driver-agnostic — the plaintext test exercises the CAS + ordering, and the E2E Failsafe path exercises TLS establishment. A dedicated TLS-carrier one-shot test would need cert+fd wiring in the carrier harness; low marginal value, noted as a follow-up.

Re-verified: transport suite **73/0 fail**, establishment test **5/5**, PMD + Checkstyle (core + community) clean.

PR #190, opened 2026-06-16
perf(transport): reuse per-stream TLS ciphertext placeholder on fd-owner ingress

arkstack-dev:
## What

fd-owner BIO `unwrap` pulls ciphertext directly from the kernel socket and **never reads its `ciphertext` argument** (`OffHeapTlsEngine#unwrap` contract, lines 396-397; impl uses `plaintext` only). Yet `NativeTcpStream.readTlsIngressFromFd` allocated a throwaway `allocateInfrastructure(1)` placeholder and `close()`d it **on every TLS record**.

This change allocates **one** empty placeholder per TLS stream and reuses it for every ingress record, releasing it deterministically in `finishCloseIfDrained()` (extracted to `releaseTlsResources()`). Only the plaintext slab is allocated per record now.

## Why — diagnosis (JFR, TLS h2load run `20260616T061640Z`)

The TLS-vs-plaintext deficit on this build is **allocation + ingress machinery (~45%)**, not crypto (FFM ~8.5%):
- `readTlsIngressFromFd` = **#1 allocation site**
- `CommunityLoanedBuffer` = **#1 allocated class**, `CommunityReleaseAction` #2

The per-read placeholder allocate+close pair was pure waste — the argument is provably ignored by fd-owner unwrap.

> Note: macro throughput/cpu_time cannot adjudicate this lever — run-to-run cpu/req variance is ±15% (0.90–1.23 ms/req across identical configs) and JFR allocation sampling is throttled (~52k cap). The justification is **No Waste Compute** + the deterministic contract, not a throughput delta.

## Test

New loopback IT (real OpenSSL handshake) **locks the invariant** this relies on: `unwrap` given a garbage, non-zero-size ciphertext buffer must still decrypt the real socket data — fails fast if a future change ever consumes the ciphertext argument.

- `OffHeapTlsEngineLoopbackIT`: 5/5 (was 4) ✅
- `NativeTcpStream*` + `NativeTcpCarrier*`: 20/20 ✅
- `pmd:check` + `checkstyle:check` (core + community): clean ✅

## Relation to #189 (drain-batch / lever A)

Supersedes #189. That branch's drain loop was confirmed a **net regression in closed-loop**: its terminating `WANT_READ` probe allocates two buffers + does an `SSL_read` *before* discovering the socket is drained, doubling per-request allocations and FFM downcalls. This PR is cut from `development/0.9.0` (fix #2), so lever A is simply not included. **#189 should be closed.**

### DIFF
--- a/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/NativeTcpStream.java
+++ b/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/NativeTcpStream.java
@@ -109,6 +109,13 @@ final class NativeTcpStream implements TransportStream {
     private final NativeTcpStreamPendingWrite.TlsContext pendingWriteTlsContext;
     private final NativeTcpStreamPendingWrite.TryWriter pendingWriteTryWriter;
 
+    // PERF: fd-owner BIO unwrap pulls ciphertext directly from the kernel socket and never
+    // inspects its {@code ciphertext} parameter (OffHeapTlsEngine#unwrap contract). A single
+    // per-stream empty placeholder is reused for every ingress record instead of allocating one
+    // per read, removing the dominant per-request LoanedBuffer/ReleaseAction churn on the TLS hot
+    // path. Owned by this stream; released in finishCloseIfDrained(). null for non-TLS streams.
+    private final LoanedBuffer tlsCiphertextPlaceholder;
+
     private final AtomicBoolean closeRequested = new AtomicBoolean(false);
     private final AtomicBoolean closed = new AtomicBoolean(false);
     private final AtomicBoolean remoteClosed = new AtomicBoolean(false);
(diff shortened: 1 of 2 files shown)


arkstack-dev:
Applied the review nits (comment-only, amended into the existing commit `57301a7e`):

1. **(Recommended)** Added a comment in `releaseTlsResources()` before `tlsCiphertextPlaceholder.close()` explaining why closing outside `tlsLock` is safe (fd-owner unwrap never dereferences the ciphertext segment) and why it must **not** be moved under `tlsLock` (held on the ingress hot path → finishCloseIfDrained must never block on it). Pre-empts a future deadlock-introducing "fix".
2. Added the `// fd-owner: serverOut stays size=0` cross-reference comment at the `wrap()` call in the new IT.

Left the `if (tlsCiphertextPlaceholder != null)` guard in place as the cautious form (per the review's note that either form is correct).

Recompile + `checkstyle:check`/`pmd:check` (core + community) clean. No behavior change.

Separately confirmed lever B works at the structural level on a fresh guided JFR run (`20260616T072510Z`): `allocateInfrastructure` dropped from **7043 → 0** appearances in allocation stacks vs the pre-B baseline (`20260616T061640Z`) — the per-record placeholder allocation is gone. As predicted, macro cpu/req is unchanged (within ±15% noise); this is a No-Waste-Compute cleanup, not a throughput lever.

PR #197, opened 2026-06-17
perf(transport): drain all buffered TLS records per readable event (ingress loop-drain)

arkstack-dev:
## Summary

Attacks the **root cause** of the TLS throughput gap, confirmed by a target-bound JFR A/B (h2 entity-read, target pinned / driver disjoint):

| | cores | rps | rps/core | CPU samples |
|---|---|---|---|---|
| plaintext h2 | 2.06 | 6477 | 3144 | 9225 |
| **TLS h2** | **5.22** | 4656 | 892 | **22370 (2.4×)** |

The TLS path burns **2.4× the CPU for fewer requests** — and it is **not** crypto (OpenSSL absent from the profile), **not** affinity (plaintext is neutral with pinning), **not** lock contention (`JavaMonitorEnter` 254; carriers idle-park on `ForkJoinPool.awaitWork`). It is **I/O event multiplication**: `NativeTcpStream.readTlsIngressFromFd()` did **one `SSL_read` (one TLS record) per reactor readable event**, so a connection with N buffered records cost N reactor iterations — each paying the full `readIngress` + slab-alloc + dispatch fixed cost. `readIngress` was the **#1 TLS hot method (12.6%)** yet absent from the plaintext top (plaintext `readPlainIngress` already drains a full slab per event). wrk never shows the gap: H1 1-req/conn = one large record, minimal events; h2load h2 multiplexes many small frames.

### Fix
`NativeTcpCarrier.readIngress()` now **loops** `readTlsIngressFromFd()` within one readable event, draining all buffered TLS records instead of one. Loop ends on `null` (WANT_READ / inbound-queue backpressure / close — all already returned as null), bounded by `MAX_TLS_RECORDS_PER_READ` (32) so one busy connection cannot starve others; the level-triggered selector re-reports any remainder next `select()`. Collapses N reactor cycles → 1.

### Scope / safety
- Community-internal, no SPI. TLS branch only (plaintext already drains a slab/event).
- Backpressure preserved (the existing `INBOUND_QUEUE_CAPACITY` guard inside `readTlsIngressFromFd` returns null when full → loop stops).
- Fairness cap + level-triggered selector → no starvation, no missed data.

## Validation
- Transport / reactor / HTTP suites green (190); real reactor TLS read path covered by `NativeTcpCarrierIngressIntegrationTest` (5 green); 0 PMD / 0 Checkstyle.
- On the target-bound rig: expect the `readIngress` CPU-profile share to drop and target rps/core to rise (2–3 reps; per-method share is the reliable signal, not single-run throughput).

Next ranked lever: egress H2 frame coalescing (#1) — batch HEADERS+DATA into fewer/larger `SSL_write` → fewer records, the egress counterpart of this fix.

### DIFF
--- a/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/NativeTcpCarrier.java
+++ b/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/NativeTcpCarrier.java
@@ -66,11 +66,11 @@ public final class NativeTcpCarrier implements TransportEngine {
     private static final String ENGINE_NAME = "CommunityNativeTcpCarrier";
     private static final int MIN_LISTENER_BACKLOG = 64;
     private static final int MAX_LISTENER_BACKLOG = 1_024;
-    // Fairness cap on TLS records drained per readable event (ingress loop-drain). One readable
-    // event pumps up to this many records (one SSL_read each) instead of one record per reactor
-    // iteration, so a single busy connection cannot monopolise the reactor thread; the
-    // level-triggered selector re-reports any remainder on the next select().
-    private static final int MAX_TLS_RECORDS_PER_READ = 32;
+    // Fairness cap on TLS records drained per readable event (see readIngress / PERF-062 in
+    // docs/subsystems/transport.md). Overridable for field tuning, mirroring
+    // exeris.transport.queueBackpressureEnabled.
+    private static final int MAX_TLS_RECORDS_PER_READ =
+            Integer.parseInt(System.getProperty("exeris.transport.maxTlsRecordsPerRead", "32"));
 
     private final TransportConfig config;
     private final MemoryAllocator allocator;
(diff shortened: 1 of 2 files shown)


arkstack-dev:
Addressed the review (commit aff3ec38):

**Requested change — TLS multi-record test (+ correcting the coverage claim).** The reviewer is correct: the description's claim that `NativeTcpCarrierIngressIntegrationTest` covers the loop-drain was **inaccurate** — that suite is plaintext (`usesFdOwnerTls()` is always false, so the TLS branch is never entered), and the existing TLS E2E sends a single record (loop exits after one iteration). Added **`NativeTcpClientServerE2eIntegrationTest.clientServerTlsRecordBurstDrains`**: a real OpenSSL TLS connection sends a burst of 16 small records; the server accumulates all of them and the test asserts every record arrives intact — exercising multiple loop-drain iterations on the TLS path. Verified locally against a self-signed cert (1 run, green); gated identically to `clientServerTlsRoundTrip` (cert / OpenSSL / `--add-opens`).

**Nits.** `MAX_TLS_RECORDS_PER_READ` is now overridable via `-Dexeris.transport.maxTlsRecordsPerRead` (default 32), mirroring `exeris.transport.queueBackpressureEnabled`. The verbose `readIngress` comment is trimmed to a one-line PERF-062 reference. The JFR cap-hit observability suggestion is deferred to a follow-up.

**Measured-effect note (transparency):** on the h2load h2 workload this dropped `readIngress` out of the hot-method list (the fix engaged) but did not move rps — the cost redistributed to the per-record buffer alloc/close cycle (~37%), which loop-drain batches the *events* of but does not reduce the *count* of. Subsequent profiling also showed the broader "TLS overhead" is largely an h2load driver-pattern artifact (wrk+TLS ≈ plaintext; Quarkus shows no h2load penalty → the residual is Exeris's per-record buffer churn vs Netty's pooling). The buffer-reuse follow-up (reused-ingress-slab) targets that.

0 PMD / 0 Checkstyle; transport/http suites green.

PR #198, opened 2026-06-17
perf(transport): reuse one ingress slab per drain — accumulate TLS records, eliminate per-record churn

arkstack-dev:
## Summary

Follow-up to #197. #197 batched the reactor **events** of TLS ingress (loop-drain) but still allocated one carrier slab — plus its `CommunityLoanedBuffer` wrapper, `CommunityReleaseAction`, and pool CLQ poll/offer — **per TLS record**. JFR alloc-by-class put that per-record machinery at **~25% of allocations** and the `close()`/`allocateBuffer`/CLQ cycle at **~37% of CPU** under fragmented (h2load) ingress.

The Quarkus cross-check was decisive: at identical H1+TLS workload, h2load made **Exeris** burn 2× the server CPU of wrk, but **Quarkus** showed *no* such penalty (Netty's pooled-buffer model carries fine-grained I/O without it). So the churn is a **real Community runtime gap**, not a driver artifact — and the segment pool (`CommunityArenaShardPool`) only pools the off-heap segment, not the heap wrapper/release-action or the per-record CLQ round-trip.

### Fix — Netty-style decrypt-into-scratch + copy
- A **per-stream reused scratch** (one 32 KB carrier slab ≥ the 16 KB max TLS record), allocated once per stream, returned to the pool in `finishCloseIfDrained()`.
- `readTlsIngressFromFd()` decrypts consecutive records into the scratch and copies each one's plaintext into a **single accumulator slab**, offering it **once**. A burst of N records costs one wrapper/release-action/CLQ round-trip instead of N (ingress requests are small, so a 32 KB accumulator holds many). TLS records are not message boundaries → concatenating plaintext is transparent to the HTTP parser.
- Accumulation stops on WANT_READ / CLOSED / before another max-size record could overflow the slab (the `readIngress` drain loop + level-triggered selector read any remainder). Per-record unwrap extracted to `unwrapNextTlsRecord()` (cognitive complexity in budget).

Trades the per-record heap-object churn for a cheap per-record memcpy — exactly Netty's tradeoff.

## Validation
- `NativeTcpClientServerE2eIntegrationTest.clientServerTlsRecordBurstDrains` (16-record burst, all bytes intact and in order) + `clientServerTlsRoundTrip` both **green over real OpenSSL TLS** — confirms accumulation correctness.
- 0 PMD / 0 Checkstyle.
- On the target-bound rig, validate by the `close()`/`allocateBuffer`/CLQ share + alloc-by-class (`CommunityLoanedBuffer` / `CommunityReleaseAction`) dropping on h2load TLS ingress.

> **Pre-existing, unrelated:** `httpSubsystemServesHealthEndpointOverTls` fails identically on `development/0.9.0` when run locally with a self-signed cert (a cert-provisioning gap in that bootstrap test, not this change).

### DIFF
--- a/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/NativeTcpStream.java
+++ b/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/NativeTcpStream.java
@@ -91,6 +91,10 @@ final class NativeTcpStream implements TransportStream {
     private static final int DRAIN_OK      = 1;
     private static final int DRAIN_STALLED = 2;
 
+    // RFC 8446 §5.1 / RFC 5246 §6.2.1: a TLS record carries at most 2^14 bytes of plaintext.
+    // The ingress accumulator stops taking records once another max-size record may not fit.
+    private static final int MAX_TLS_RECORD_PLAINTEXT_BYTES = 16 * 1_024;
+
     private static final int INBOUND_ADVANCE_NEEDS_MORE_DATA = -1;
     private static final int INBOUND_ADVANCE_RETRY = 0;
     private static final int INBOUND_ADVANCE_READY = 1;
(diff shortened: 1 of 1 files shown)


arkstack-dev:
Closing — measured net-negative. Reused-slab accumulation engaged (ArenaShardPool.Allocation 2101→84) but did not move target-bound throughput and regressed: it added a per-record scratch→accumulator memcpy (readTlsIngressFromFd back to #1 hot method at 12.55%, vs ~0 after #197), Young GC stayed high (869), and peak RSS rose (~540→918MB). Root reason: ingress was already handled by #197's loop-drain; the dominant remaining buffer churn is EGRESS (response-frame writes: CLQ poll/offer ~15.6% + close + allocateBuffer), which this PR doesn't touch. Pivoting to egress coalescing / wrapper-object pooling. The TLS multi-record burst test added here is worth salvaging separately.

PR #199, opened 2026-06-17
perf(http2): coalesce a response's frames into one socket write (egress)

arkstack-dev:
## Summary

The egress counterpart to #197's ingress loop-drain, targeting the **dominant remaining TLS churn**. Per the #198 analysis, the per-record buffer cycle (CLQ poll/offer ~15.6% + `close()` + `allocateBuffer`) is **egress-bound**: `writeHttp2Response` issued a separate `stream.write()` for the HEADERS frame and for **each** DATA frame, so every response cost ≥2 outbound enqueues — and on TLS, ≥2 `PendingWrite` + `wrap()` + `SSL_write` cycles.

### Fix
Serialize the whole framed response (HEADERS[+CONTINUATION] + all DATA) into **one** buffer (sized to the framed length via `frameCount()`) and issue a **single** `stream.write()`. The wire framing is **unchanged** — still multiple H2 frames, each ≤ `SETTINGS_MAX_FRAME_SIZE`; only the socket write is coalesced, so the TLS path does one `wrap()`/`SSL_write` and the reactor one outbound enqueue per response.

The per-frame writers (`writeHttp2HeaderBlockFrames` / `writeHttp2DataFrames` / `writeHttp2Frame`) become offset-based serialize helpers (`serializeHeaderBlockFrames` / `serializeDataFrames` / `serializeFrame`) that write into the shared buffer and return the advanced position.

### Why this one (vs the closed #198)
#198 (reused ingress slab) was net-negative — ingress was already handled by #197, and it added a memcpy + RSS. Egress is where the response-side per-frame churn actually lives, and the Quarkus cross-check pinned the gap to fragmented-I/O buffer churn. This attacks that directly.

## Tests
- New `h2ResponseHeadersAndDataCoalescedIntoOneWrite`: a body response's HEADERS and DATA frames arrive in **exactly one** socket write (`CapturingTransportStream` now records per-write segments), and the DATA payload is wire-correct.
- Full HTTP/2 codec suite green (framing unchanged — the response-decoding tests confirm wire correctness).
- 0 PMD / 0 Checkstyle.

## Validation on the target-bound rig
Re-bench h2load H2+TLS: expect the egress `stream.write` count to drop to ~1/response and the CLQ poll/offer + `close`/`allocateBuffer` egress share to fall (and, if the box is target-CPU-bound, rps/core to rise). Take 2–3 reps.

### DIFF
--- a/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/http/CommunityHttp2SessionProcessor.java
+++ b/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/http/CommunityHttp2SessionProcessor.java
@@ -428,17 +428,20 @@ final class CommunityHttp2SessionProcessor {
                         position = serializeDataFrames(
                                 outbound.segment(), position, streamId, bodyBuffer.segment(), bodyBytes);
                     }
-                    outbound.setSize(position);
                     stream.write(outbound.segment(), (int) position);
                 }
             }
         }
     }
 
-    /** Number of frames a payload of {@code payloadLength} bytes spans (one frame even when empty). */
+    /**
+     * Number of HTTP/2 frames a payload of {@code payloadLength} bytes spans — zero when empty, to
+     * match the serialize loops (which emit no frame for a zero-length payload). Used only to size
+     * the coalescing buffer; the actual written length is the position returned by serialization.
+     */
     private static int frameCount(int payloadLength) {
         if (payloadLength <= 0) {
-            return 1;
+            return 0;
         }
         return (payloadLength + HTTP2_MAX_FRAME_PAYLOAD_BYTES - 1) / HTTP2_MAX_FRAME_PAYLOAD_BYTES;
     }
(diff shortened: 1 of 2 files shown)


arkstack-dev:
Addressed review feedback and rebased onto `development/0.9.0` (the branch was behind #200/#201; the rebase also resolves the diagnostics-SPI skew from #201 so it builds against current base).

**Review items:**
- **Nit — `frameCount(0)` semantics:** now returns `0` (not `1`) for an empty payload, matching the serialize loops (which emit no frame for zero length). It only sizes the coalescing buffer and `headerBytes` is always `> 0`, so wire output is unchanged; this just removes the HEADERS-vs-loop inconsistency.
- **Nit — redundant `outbound.setSize(position)`:** dropped. `stream.write(segment, (int) position)` takes the explicit length and the buffer closes immediately after, so the logical size was unused (matches the sibling raw-write path).
- **Main item — multi-DATA-frame test:** added `h2MultiDataFrameBodyCoalescedAndWireCorrect`. A 40000-byte body spans 3 DATA frames; asserts all HEADERS+DATA arrive in **one** socket write, each frame ≤ 16 KiB, `END_STREAM` on the final DATA frame only, and the reassembled (position-dependent) payload equals the body — guarding the `serializeDataFrames` offset arithmetic.
- The ≥2 GB `(int) bodyBuffer.size()` cast is pre-existing and out of scope (no response-body-size limit in this path), left as-is.

Verified: `CommunityHttpRequestProcessorTest` 17/17 green, PMD + Checkstyle clean.

Issue #180, opened 2026-06-11
AbstractTransportStreamTck.ResetContract: add drain-vs-abandon discriminator (reset must not be satisfiable by graceful close)


arkstack-dev:
# AbstractTransportStreamTck.ResetContract cannot discriminate abandon (reset) from drain (close)

## Problem

`ResetContract` is the declared TCK-level replacement for runtime abort guards in
downstream handlers (enterprise PR #39 removed the `instanceof AbortableStream`
runtime check on that basis). In its current shape the suite passes for the SPI
**default** implementation (`reset(long)` → graceful `close()`):

- `resetAbandonsPendingData` polls `hasPendingData() == false` after `reset(0L)` —
  but a graceful close that *drains* queued writes also ends with
  `hasPendingData() == false`. Abandon and drain are indistinguishable to the
  assertion. (Bindings without a live carrier flush synchronously inside
  `queueWrite`, making the case additionally vacuous-by-timing.)
- `readAfterResetThrows` / `writeAfterResetThrows` / `queueWriteAfterResetThrows` /
  `resetIsIdempotentAndComposesWithClose` are all satisfied by a plain `close()`
  after the read-on-locally-closed contract (kernel #177/#178 line).

Net effect: a transport that silently degrades abort to graceful close fully
passes `ResetContract`. The only tests in the ecosystem that actually pin abort
semantics are per-carrier unit pins (e.g. enterprise
`QuicTransportStreamWriteOnClosedTest.reset_mapsErrorCodeToNativeResetStreamAndCloses`).

## Proposal

Add a drain-vs-abandon discriminator to `ResetContract`, e.g.:

1. A TCK seam reporting whether queued-but-unflushed data was **delivered** to the
   peer after `reset()` — abandon means the peer MUST NOT observe the queued
   payload (vs `close()`, which SHOULD drain it). Requires a `StreamPair` harness
   capable of holding egress back (backpressure gate) so data is genuinely pending
   at reset time.
2. Alternatively/additionally: an optional `expectsTrueReset()` flag — when true,
   the binding must provide an observer for the transport's native reset signal
   (error code visible at the peer), turning the mapping pin into a TCK-level
   obligation instead of a per-repo convention.

## Context

- Enterprise: `docs/subsystems/transport.md` conformance rule now states the
  two-part requirement (bind `AbstractTransportStreamTck` + per-carrier reset
  mapping test) and links this issue.
[...]

PR #202, opened 2026-06-18
fix(transport): abortively cancel the key in closeKeyStream to stop closed-engine TLS reactor spin

arkstack-dev:
## Problem

Under HTTP/2 over TLS the Community `NativeTcp` transport burned **2–2.4× the CPU per request** of the same workload without TLS (and of Quarkus over TLS), while throughput looked fine. The deficit was pure reactor overhead, not crypto.

## Root cause (JFR-confirmed)

The reactor select loop reaches `NativeTcpCarrier.closeKeyStream` only via its `catch (RuntimeException)` branch — a per-key dispatch (read-ingress or flush) faulted. The dominant trigger under TLS h2 is an **unwrap-on-closed `TlsDecryptException`** (the pre-allocated `DECRYPT_CLOSED_SENTINEL` in `OffHeapTlsEngine`) thrown once a stream's TLS engine is already closed but its graceful close is still deferred for outbound drain (`NativeTcpStream#finishCloseIfDrained` returns early while the outbound consumer slot is owned by another thread or data is pending).

Because the deferred graceful close never reaches `channel.close()`, the `SelectionKey` stays registered; the level-triggered channel re-fires every select, re-throwing and **busy-spinning the reactor** — measured **~688M throws / 120s (~5.7M/s, ~47% CPU)**. It was invisible to JFR exception sampling because the sentinel is zero-alloc with no stack trace (the earlier "~200 exceptions" reading was a sampling artifact).

Withdrawing only `OP_READ` is insufficient: a closed-engine stream's queued egress is **undrainable** (cannot wrap), so `flushStream` never narrows interest and the spin merely relocates onto `OP_WRITE` (measured: `flushStream` jumps to ~23% CPU, `cpu_time` unchanged).

## Fix

`closeKeyStream` is the reactor's error-teardown path, so tear the connection down **abortively**:

1. **`key.cancel()` synchronously** — removes the key from this reactor's selector so the dead channel cannot be re-selected for read **or** write. Direct cancel honours the single-consumer key protocol (it runs on the reactor thread).
2. **`reset()` instead of `close()`** — `reset` abandons the undrainable queue and sets `resetRequested`, the designed escape hatch (issue #180) that lets `finishCloseIfDrained` bypass its drain gate, finalize, and deregister the channel, including under slot-owner deferral via the deferred-abort wake. A graceful `close()` would defer forever here.

Abortive-on-error matches the existing `abortOnUnrecoverableWriteFailure` pattern. **Core `OffHeapTlsEngine` is unchanged** — the sentinel is the correct terminal signal; the defect was the transport treating it as retryable.

## Results (live benchmark, entity-read-by-id, TLS h2, loopback, 120s)

| Metric | Before | After |
|---|---|---|
| Throughput | ~4,100 rps | **6,770 rps** (+65%) |
| Target-JVM CPU time | ~680 s | **332 s** (−51%) |
| Efficiency | ~723 rps/core | **~2,444 rps/core** (~3.4×) |
| h2load errors | 0 | **0** |

That lifts TLS h2 efficiency to ~h2c plaintext (~2,790 rps/core) — the TLS CPU/req penalty is eliminated. JFR hot-methods after the fix are entirely legitimate application work (Postgres JDBC + Jackson + FFM `MethodHandle`); the whole transport spin chain (`flushStream`, `finishCloseIfDrained`, `runLoop`, `closeKeyStream`, `AbstractLoanedBuffer.close`, `CLQ.poll/offer`) is gone from the hot set — eliminated, not relocated.

## Tests

Adds `NativeTcpCloseKeyStreamReadInterestTest` (community impl-level): builds a carrier + selector + loopback channel armed for `OP_READ`, calls `closeKeyStream`, asserts the key is cancelled synchronously. Verified **fails before the fix** (key stays valid) and **passes after**. Existing transport suite green (41 tests incl. multi-reactor and the #180 deferred-abort TCK).

## Scope / boundary

Community-only. **No SPI / Abstract*Tck / ADR change** — selector mechanics are not an SPI contract (ADR-008: high-perf transport is Enterprise; SPI `TransportStream` is implementation-blind). Reviewed via `exeris-architect` (placement + abortive-teardown re-validation) and `exeris-tck` (test contract).

### DIFF
--- a/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/CommunityReactorDispatchFaultEvent.java
+++ b/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/transport/CommunityReactorDispatchFaultEvent.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2025-2026 Exeris Systems.
+ *
+ * Licensed under the Apache License, Version 2.0 with Commons Clause.
+ * You may use, modify, and distribute this file under those terms.
+ * Commercial resale of this software as a competing product is prohibited.
+ * See LICENSE-COMMUNITY in the repository root for the full text.
+ */
+package eu.exeris.kernel.community.transport;
+
+import jdk.jfr.Category;
+import jdk.jfr.Description;
+import jdk.jfr.Event;
+import jdk.jfr.FlightRecorder;
+import jdk.jfr.Label;
+import jdk.jfr.Name;
+import jdk.jfr.StackTrace;
+
+/**
+ * JFR event for an unrecoverable per-stream fault in the reactor select loop.
+ *
+ * <h2>JFR-First Contract</h2>
+ * <p>Emitted once per dying connection, from the {@code NativeTcpReactor} select-loop
+ * {@code catch (RuntimeException)} branch that triggers {@link NativeTcpCarrier#closeKeyStream}
+ * abortive teardown. This is a failure-detection point (per {@code docs/subsystems/transport.md}
+ * JFR-first guidance): a recurring fault here is the signature of a reactor spin, which is otherwise
+ * invisible to JFR exception sampling when the cause is a zero-alloc, stack-trace-less sentinel
+ * (e.g. unwrap-on-closed {@code TlsDecryptException}). The payload is the exception <em>class</em>
+ * name only — no message — so it carries no request data.
+ *
+ * <p>Single-phase {@code commit()} on the reactor's platform thread; zero overhead when JFR is not
+ * recording ({@link #isEnabled()} check). Not on the read/write ingress hot path.
+ *
+ * @since 0.9.0
+ */
+@Name("eu.exeris.kernel.transport.CommunityReactorDispatchFault")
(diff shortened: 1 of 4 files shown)


arkstack-dev:
Addressed the two non-blocking review items:

- **JFR-first failure-detection event** (the main suggestion): added `CommunityReactorDispatchFaultEvent`, emitted from the `NativeTcpReactor` select-loop `catch (RuntimeException)` branch — one per dying connection, off the ingress hot path. Payload is `streamId` + `reactorIndex` + the exception **class** name only (no message — secret-safe). A recurring fault here is the spin signature that was invisible to JFR exception sampling because the cause is a zero-alloc, stack-trace-less sentinel — this closes that blind spot. Single-phase `commit()` on the reactor's platform thread, zero overhead when not recording.
- **Cosmetic test nit:** added the comment clarifying `serverChannel` is accepted only to complete the TCP handshake, never read.

Verified: transport tests green (23 incl. the regression + close/reactor suite), PMD + Checkstyle clean.

PR #160, opened 2026-06-03
Release v0.8.0 — production-correctness + HTTP body-codec SPI matrix + RBAC wiring

arkstack-dev:
# Release v0.8.0

Cut of `development/0.8.0` → `main`. 42 commits, Sprints 1–8. Full details: [`docs/release/v0.8.0-release-notes.md`](docs/release/v0.8.0-release-notes.md); terse summary in [`CHANGELOG.md`](CHANGELOG.md) `[0.8.0] — 2026-06-03`.

## Three architectural arcs
1. **HTTP body-codec SPI matrix — fourth seam.** Server-side request-body **decode** (`HttpRequestBodyDecoder` + registry + context + `CommunityJsonRequestBodyDecoder` + TCK, ADR-036/HTTP-138) completes the SPI surface across all four `{request,response}×{encode,decode}` quadrants. Tier-neutral `KernelWebClient` facade in Core + `HttpClientRequestEnricher` (ADR-034 / ADR-032). *The generator that consumes the new server-side decoder (exeris-tooling TOOL-139) is snapshot-gated and not in this release — generated handlers still inline-decode until that later cycle.*
2. **Compile-time RBAC → live runtime decision.** `@RequiresRole` APT registry wired into the runtime (`GeneratedRoleRegistryLoader` + `roleMask`), fail-closed by construction (SEC-080, ADR-014).
3. **Transport / Flow / Graph → production-candidate confidence.** Non-blocking reactor-driven client ingress removes VT-carrier pinning under constrained cores (TCK-064); GRAPH-111 + FLOW-110 add mandatory TCK gates.

Also discharges the v0.7 review tail (QA-010..018 decompositions, `kafka-clients` 4.0.0, JaCoCo + Kafka-IT CI gates).

## The Wall held throughout
No Jackson, driver, or native types in any SPI signature; codec drivers keep `ObjectMapper` confined to `exeris-kernel-community`.

## Honest carry-over to v0.9 (not in this release)
`UriTemplate`/`QueryParams` (HTTP-130/131) + `CommunityKernelContextEnricher` (HTTP-134) did not land; `KernelDiagnostics` (ADR-033) is decision-only; lockstep TOOL-139 + Enterprise stub are snapshot-gated; IdentityProvider SPI RFC; FLOW-110 `closeTimeoutNanos` follow-up; durability JMH (exeris-benchmarks). All annotated in ROADMAP (`Status (v0.8)`: DELIVERED / PARTIAL / CARRIED TO v0.9).

## Version
[...]

PR #208, opened 2026-06-19
fix(flow): fence stale-runtime PARK persists from resurrecting reclaimed saga rows

arkstack-dev:
## Fix a load-sensitive saga-recovery reclaim race (CI-only flake)

`AbstractSagaRecoveryTck$RestartUnderLoad.parkedFleetSurvivesForceCloseAndResumes` flakes on 2-vCPU CI runners (green on many-core dev boxes). Root cause is a real Core ordering defect, **pre-existing since v0.8.0 (#160)** — not introduced by v0.9.

### Mechanism
A flow worker abandoned past `close()`'s bounded interrupt+join (`CoreFlowRuntime.interruptAndJoinRunningThreads`, 5s/thread) belongs to a now-**closed** runtime. In the multi-runtime restart model (a fresh `CoreFlowRuntime` rebuilt on the same shared `FlowSnapshotStore` — both production restart and the saga-recovery TCK do this), that straggler can land a `save(PARKED)` **after** the rebuilt runtime drove the same flow key to `COMPLETED` and reclaimed (deleted) the row via `complete()` → `deleteSnapshot()`. The gen-1 `PARKED@step1` row is resurrected, breaking the unbounded-mode "deleted on `complete()`" invariant (`docs/subsystems/flow.md`). On a 2-vCPU runner the scheduling window that lets the straggler outlive the join is wide; on many cores it effectively never opens.

### Fix
A lifecycle write-fence in `CoreFlowRuntime.persistSnapshot`: skip the save when `!isActiveLifecycle(instance)`.

`isStaleLifecycle` alone is **insufficient** — the abandoned worker's own generation counter is unchanged by the *other* runtime's `start()`, so it never looks "stale" relative to its own (closed) runtime. The guard must also reject a closed/not-started runtime (`started && !closed`). This is the stronger sibling of the `staleLifecycle`/`cleanupOnly` guards already on `deleteSnapshot` and `finalizeFailedInstance`. Verified not to over-block: an active engine still persists PARK normally.

### Verification
- New deterministic regression `CoreFlowRuntimeStaleWriteFenceTest` (two runtimes / one shared store / latch-pinned gen-1 worker): `savesAfterReclaim=1` (row resurrected) with the old `isStaleLifecycle` predicate, **0 with the fence**. A sanity method confirms an active engine still persists PARK.
- `mvn pmd:check checkstyle:check -pl exeris-kernel-core` → 0 violations, BUILD SUCCESS.
- The flaky `RestartUnderLoad` could not be reproduced locally even at `-Djdk.virtualThreadScheduler.parallelism=1` (0/18) — confirming the race is environment/load-sensitive; hence the deterministic two-runtime detector rather than relying on the fleet test.

### Scope
SPI-blind, Core-only, allocation-free hot-path-neutral guard (one predicate). No SPI/TCK signature change. Targets `development/0.9.0` so it rides #207 into `main` before the `v0.9.0` tag.

Reviewed via exeris-architect (root cause) and exeris-tck (which caught that the first predicate `isStaleLifecycle` did **not** fix the multi-runtime case and proved `!isActiveLifecycle` does).

🤖 Generated with Claude Code


arkstack-dev:
Refinement after a local full-suite run: the first predicate (`!isActiveLifecycle` alone) over-blocked and broke `CoreFlowRuntimeTest.lateFailedExitAfterCloseStillReleasesClaimsAndFinalizesSnapshot` — a deliberate cleanup contract where a **late FAIL after `close()` must still persist `FAILED_ROLLEDBACK`** on the same runtime. 

Narrowed the fence to `!isActiveLifecycle(instance) && !state.isTerminal()`: terminal finalizations are exempt (a terminal state is not a resurrected checkpoint), only non-terminal PARK re-checkpoints from a non-active runtime are fenced. 

Verified: `CoreFlowRuntimeStaleWriteFenceTest` 2/2, `CoreFlowRuntimeTest` 20/20 (late-fail contract green again), flaky `immediateScheduleParkWakeOnSameContextIsRaceSafe` 3/3 (unrelated pre-existing flake, not worsened), `pmd:check`+`checkstyle:check` 0 violations.

arkstack-dev:
Addressed both reviews' non-blocking nits (comment/doc-only, no behavior change):
- `flow.md` predicate citation updated to `!isActiveLifecycle && !isTerminal`.
- Inline comment in `persistSnapshot` now notes the two-runtime late-FAIL orphan implication (benign — `terminalStateCatalog` records COMPLETED, no re-execution).
- `liveGen1WorkerCount` comment clarifies the shared thread-name and the gen-2-completed-first dependency.

Re-verified: `CoreFlowRuntimeStaleWriteFenceTest` 2/2, `CoreFlowRuntimeTest` 20/20, `pmd:check`+`checkstyle:check` 0 violations.

PR #209, opened 2026-06-19
chore(v0.9): clear PR #207 review nits + CodeQL findings

arkstack-dev:
## Clear PR #207 review nits + CodeQL findings

Low-severity, behavior-neutral follow-ups from the v0.9.0 milestone review (#207). Targets `development/0.9.0` so it rides into `main` via #207 before the `v0.9.0` tag.

### CodeQL / code-quality bot
- **`NativeTcpCarrier` — uncaught `NumberFormatException` in static init.** `MAX_TLS_RECORDS_PER_READ` parsed `-Dexeris.transport.maxTlsRecordsPerRead` directly in the field initializer; a malformed flag would fail class load. Extracted `readMaxTlsRecordsPerRead()` with a fallback to the documented default (`32`) + a `WARNING` log. (Helper placed with the other `private static` helpers to keep PMD `FieldDeclarationsShouldBeAtStartOfClass` happy.)
- **`TransportStream.reset(long)` — "useless parameter".** The default impl deliberately discards the advisory `errorCode` and delegates to `close()`. Acknowledged via the unnamed-variable idiom (`long _ = errorCode;`) — keeps the SPI contract parameter, signals intent, no broad suppression.

### Main review nits
- **`NativeTcpStream.lastQueueDepth` — needless `volatile`.** Read/written only on the reactor thread (`offerIngress`) and feeds an advisory JFR "trend" label; dropped to a plain `int` (No Waste Compute — removes a per-record fence). Comment documents the single-thread invariant.
- **`CommunityJwksKeyRotationEvent.newKid` — permanently-null JFR column.** `null` at all three call sites: JWKS rotation adopts a key *set*, not a single new kid, and `currentKeyCount`/`previousKeyCount` already carry the set sizes. Removed the field, updated the 3 `emit` sites and the JFR field-surface assertion. `retiredKid` (populated on `CUTOVER_DENY`) is kept. The event is new in v0.9 and not yet released, so removing the dead column now avoids ever shipping it.

### Informational
- **ADR-039 ArchUnit obligation — already present.** `ExerisArchitectureTest.diagnosticsSpiIsEventFree` bans `eu.exeris.kernel.spi.diagnostics..` from depending on `jdk.jfr..` / `eu.exeris.telemetry.spec..`, citing ADR-033 Obligation 10 / ADR-039. No `@ArchIgnore`; the suite is active (10/10 green). The v0.8 "ArchTest ignored" finding was resolved during v0.9.

### Verification
`CommunityJwksKeyRotationJfrTest`, `NativeTcpCloseKeyStreamReadInterestTest`, `ExerisArchitectureTest` (10/10) green; `pmd:check` + `checkstyle:check` → 0 violations on `spi` + `community`.

🤖 Generated with Claude Code

### DIFF
--- a/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/security/CommunityJwksKeyRotationEvent.java
+++ b/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/security/CommunityJwksKeyRotationEvent.java
@@ -42,9 +42,6 @@ final class CommunityJwksKeyRotationEvent extends Event {
     @Label("Phase")
     /* default */ String phase;
 
-    @Label("New Kid")
-    /* default */ String newKid;
-
     @Label("Retired Kid")
     /* default */ String retiredKid;
 
@@ -54,7 +51,7 @@ final class CommunityJwksKeyRotationEvent extends Event {
     @Label("Previous Key Count")
     /* default */ int previousKeyCount;
 
-    /* default */ static void emit(String phase, String newKid, String retiredKid,
+    /* default */ static void emit(String phase, String retiredKid,
                                    int currentKeyCount, int previousKeyCount) {
         if (!FlightRecorder.isInitialized()) {
             return;
@@ -62,7 +59,6 @@ final class CommunityJwksKeyRotationEvent extends Event {
         CommunityJwksKeyRotationEvent event = new CommunityJwksKeyRotationEvent();
         if (event.isEnabled()) {
             event.phase = phase;
-            event.newKid = newKid;
             event.retiredKid = retiredKid;
             event.currentKeyCount = currentKeyCount;
             event.previousKeyCount = previousKeyCount;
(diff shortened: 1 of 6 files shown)


arkstack-dev:
Applied the range-validation suggestion (non-blocking #1): `readMaxTlsRecordsPerRead()` now rejects non-positive values too — a `-Dexeris.transport.maxTlsRecordsPerRead=0` would have silently disabled per-readable TLS record draining for every connection on the carrier (the drain loop never iterates). Falls back to the default 32 with a WARNING, same as the malformed-value path. Re-verified: `pmd:check`+`checkstyle:check` 0 violations.

Skipped suggestion #2 (`.doesNotContain("newKid")`): as noted, the field is gone from the class so its absence is compile-enforced — a runtime assertion on a non-existent field would be noise.

arkstack-dev:
Added a `### Fixed` entry to CHANGELOG `[0.9.0]` for #208 (saga recovery stale-write fence) — folding in the PR #207 conditional-approval doc note here since this branch is the v0.9 release-hygiene vehicle. Separate doc-only commit; no code change to the already-reviewed nits.

PR #212, opened 2026-06-19
fix(spi): drop the CodeQL read-and-discard silencer in TransportStream.reset default

arkstack-dev:
Breaks a CodeQL tail-chase introduced in #209.

**The loop:**
- bare `close()` → CodeQL "useless parameter" (`errorCode` unused)
- `long _ = errorCode;` (added in #209) → CodeQL "unread local variable" on the discard

Every code state trips one Note-level false positive. `errorCode` is genuine SPI contract for overriders (`NativeTcpStream.reset` maps it to a protocol RST via SO_LINGER 0); the default method has no abort primitive and intentionally discards it, delegating to `close()`.

**Resolution:** clean default body (no ceremony) + **dismiss** the "useless parameter" alert as a known false positive on an interface default method whose param exists for overriders. The comment now explicitly forbids re-adding a read-and-discard line, so the loop cannot reopen.

> Action for the maintainer: dismiss the CodeQL `java/useless-parameter` alert on `TransportStream.reset` as a false positive (interface default-method contract param). Do not re-silence in code.

SPI-only, no behavior change, no boundary impact. `pmd:check` + `checkstyle:check` on `exeris-kernel-spi` clean.

🤖 Generated with Claude Code

### DIFF
--- a/exeris-kernel-spi/src/main/java/eu/exeris/kernel/spi/transport/TransportStream.java
+++ b/exeris-kernel-spi/src/main/java/eu/exeris/kernel/spi/transport/TransportStream.java
@@ -210,8 +210,11 @@ public interface TransportStream extends AutoCloseable {
      * @param errorCode caller-supplied protocol error code (advisory; transport-mapped)
      */
     default void reset(long errorCode) {
-        // The default has no abort primitive and discards the advisory code; overrides map it.
-        long _ = errorCode;
+        // errorCode is contract for overriders (e.g. a TCP driver maps it to a protocol RST via
+        // SO_LINGER 0); the default has no abort primitive and intentionally discards it. CodeQL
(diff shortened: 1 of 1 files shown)


arkstack-dev:
Applied the non-blocking Wall-purity note: the inline comment no longer names the concrete `NativeTcpStream` impl class — genericised to "a TCP driver maps it to a protocol RST via SO_LINGER 0", keeping the concrete example and the loop-prevention guidance intact while keeping the SPI source implementation-blind. Checkstyle on `exeris-kernel-spi` clean.

Reminder of the mandatory companion action (cannot be done in code): **dismiss the CodeQL `java/useless-parameter` alert on `TransportStream.reset`** as a false positive (interface default-method contract parameter).

PR #213, opened 2026-06-20
docs(roadmap): correct Platform Baseline — preview branch, not STS adapter artifact

arkstack-dev:
Corrects the **Platform Baseline for 1.0 GA** section (added in #211) so it does not mislead before #207 carries it to `main`.

## Why
Two facts settle the preview-decouple design differently than #211 stated:
1. **STS API moves per JDK release** and our code targets **JDK 26**, not 25 → the STS variant must track a moving preview API on a separate JDK/build line, which a **branch** absorbs naturally and a same-reactor adapter module does not (`--enable-preview` pins bytecode to one major).
2. **The bootstrap already exposes an orchestration seam** → `main` carries only a *GA-clean substitution* at that seam (never two implementations side by side), and the STS / value-class version belongs on a **thin temporary `preview` branch** — the *intended future `main`* — not an `exeris-kernel-sts` adapter artifact.

## What changed (docs-only)
- **Step 2** — reframed from "own VT layer, a feature not a compromise" (permanent default) to a **temporary GA-clean substitution at the existing seam**, on `main` only; STS stays the intended impl.
- **Step 3** — replaced the "opt-in adapter artifact (`exeris-kernel-sts`)" with the **`preview` branch model**: `preview` = richer target kernel (STS + later JEP 401 value carriers) that **replaces `main`** when those reach GA; `main` = preview-clean cut. Direction corrected (preview is not an overlay added to main).
- **Merge gate** — preview-bearing code is **never on `main`**, so there is no adapter dependency to exclude; the gate is just "zero preview bytecode on `main`".
- **CLAUDE.md note + Strong-Defaults guardrail + summary bullet + ADR note** — updated in lockstep (STS scoped to JVM-controlled deployments + the preview branch).

No behavioural/code change; forward-looking roadmap intent only.


arkstack-dev:
Addressed both non-blocking nits in `5b51bb23`:

1. **CLAUDE.md mixed registers** — rescoped to one concept: `scoped to JVM-controlled deployments (served via the `1.0-preview` artifact on the `preview` branch, the intended future `main`)`. The branch is now surfaced as *how* those deployments are served, matching step 3's wording.
2. **Step 2 seam precision** — named the concrete boundary: `at the existing concurrency-policy method boundary (e.g. the private `startParallel` in `SubsystemOrchestrator`) — not a new injectable strategy`. Pre-empts any pluggable-interface misread.

Thanks for the verification on the `--enable-preview` major-pinning / per-JDK STS-API-churn point and the `startParallel` boundary — both confirm the corrected model.

PR #215, opened 2026-06-22
fix(flow): fail-closed SCHEMA_MISMATCH on version-blind saga resume (v0.10 S11)

arkstack-dev:
## Summary

v0.10 Sprint 11 — closes a **data-corruption-class** gap in saga resume. A parked saga resuming under a *redeployed* flow definition replayed its persisted step index with **no bounds check**: a removed/shrunk definition made the stale `currentStep` silently bind to a *different* step. The `EX-FLOW-7002 / phase=SCHEMA_MISMATCH` behaviour `docs/subsystems/flow.md` already promised **did not exist** — only a name-mismatch check did.

## Change

- **`CoreFlowRuntime.validateSnapshotStepBounds`** — on resume, a non-terminal persisted `currentStep` that no longer indexes a step in the resolved plan throws **fail-closed before any VT launch / step replay** (synchronous on `wake()`). Terminal snapshots are exempt (never resumed). Bounds/arity guard — full step-identity validation (same-arity reorders) is the v0.11 definition-versioning epic.
- **`FlowEngineException.schemaMismatch(engineName, persistedStep)`** — Glass-Box factory: `phase=SCHEMA_MISMATCH`, `reason=STEP_OUT_OF_RANGE`, `contextValue=step`, reusing `EX-FLOW-7002` with the **existing rawArgs schema** (one-code-one-schema).
- **Tests** — `CoreFlowRuntimeTest` unit case (park tail step → redeploy a shorter definition → `wake()` fails closed) + `AbstractSagaRecoveryTck.SchemaMismatchOnResume` cross-binding case (green on **Core + Community**), both asserting the structured rawArgs (phase + out-of-range step) and that **no step re-executes** on the rejected resume.
- **`flow.md`** — drift corrected: the removal row is now enforced (bounds/arity); the same-arity reorder row is honestly noted as **not yet caught** (v0.11).

## Scope

Stage 2 (`FlowDefinition` name+version key + in-flight vN→vN+1 migration) stays **v0.11**. `spi/core/tck` `pmd:check` + `checkstyle:check` clean.


arkstack-dev:
Addressed the review in `06e9a0f9` (force-pushed). All four findings handled:

- **#1 — `lookupParked` can now throw (contract change).** Added a `@throws FlowEngineException (phase=SCHEMA_MISMATCH / EX-FLOW-7002)` clause to `FlowScheduler.lookupParked` Javadoc, noting the snapshot-fallback path can surface it (fail-loud is the defensively-correct choice for a data-corruption-class outcome) while the in-memory fast path never does.
- **#2 — Missing JFR event (was tracked-follow-up; landed now).** Added `FlowSchemaMismatchEvent` (`eu.exeris.kernel.flow.SchemaMismatch`, single-phase, `@StackTrace(false)`), emitted at the guard's single throw site (`validateSnapshotStepBounds`) — operators detect deployment-gone-wrong via JFR, not log-scraping. Secret-safe (engine/definition names + step counters only). Documented in the flow.md JFR table + verified by a `RecordingStream` test in `CoreFlowRuntimeTest`.
- **#3 — `step < 0` not guarded.** Added `step < 0 || step >= stepCount` so a corrupted sentinel index also fails closed; invariant now explicit.
- **#4 — TCK does not assert `rawArgs[2]`.** Both the `AbstractSagaRecoveryTck.SchemaMismatchOnResume` case and the `CoreFlowRuntimeTest` unit case now assert `rawArgs[2] == "STEP_OUT_OF_RANGE"` (the stable reason key) alongside phase + step.

Verification: `CoreFlowRuntimeTest` 22/22 (incl. the new JFR-emission test), `AbstractSagaRecoveryTck.SchemaMismatchOnResume` green on Core + Community, `pmd:check` + `checkstyle:check` clean on spi/core/tck/community.

Scope unchanged — Stage 2 (definition versioning + in-flight migration) stays v0.11.

arkstack-dev:
**SonarQube new-code duplication** (FlowSchemaMismatchEvent 22.2%) resolved in `f98a3761`.

Root cause: the new JFR event class structurally mirrors the existing `FlowTimeoutEvent` (shared `engineName`/`definitionName`/`instanceId*` `@Label` fields + the `emit()` `isEnabled→begin→set→commit` skeleton). This is **framework-mandated JFR boilerplate**, not a DRY violation — every `jdk.jfr.Event` must be its own concrete class with its own `@Name`, so the field/emit shape repeats by design. **75 of 77 `*Event.java`** in the kernel extend `jdk.jfr.Event`.

Fix: added `**/*Event.java` to `sonar.cpd.exclusions` — the same treatment the repo already gives `*Benchmark*` harness classes (boilerplate where copy-paste detection is noise). The only two non-JFR `*Event.java` (`KernelEvent` base, `StreamEvent` SSE record) are tiny carriers, not CPD sources, so the exclusion has no real coverage cost. Comment in `sonar-project.properties` documents the rationale.

Alternative considered and rejected: an `AbstractFlowEvent` base to share fields — the flow events don't share a common field set (`WakeOnLoadFallbackEvent`/`OptimisticLockConflictEvent` carry different fields), and JFR-event inheritance refactoring of already-shipped events is out of scope for this fix.

PR #241, opened 2026-07-19
fix(http): stream JSON response encoding into pooled LoanedBuffer

arkstack-dev:
## Problem
`JsonBodyEncoder.encode` serialized the response payload to a heap `byte[]` via
`ObjectMapper.writeValueAsBytes`, then `MemorySegment.copy`'d it into the loaned,
registered network buffer. That is a per-response heap allocation **plus** a
full extra pass over the payload — the copy exists only to bridge Jackson's heap output
into the off-heap response buffer. In a perfbox JFR pair this response-encode path was a
large share of allocation (growable `ByteArrayBuilder` + `toByteArray` + the copy).

## Fix
Jackson now streams straight into the loaned segment through a new `SegmentSink`
(an `OutputStream` over a `LoanedBuffer` with grow-on-overflow), so `writeValue` writes
directly off-heap. Both the intermediate `byte[]` and the `MemorySegment.copy` are gone.

- **Output is byte-identical**; **no SPI change** — this is an internal reimplementation
  of one Community encoder, so The Wall (ADR-006) holds untouched.
- `SegmentSink` keeps **at most one live buffer**: on overflow it allocates a larger one,
  copies, closes the old. `close()` is a no-op (a serializer may auto-close its target);
  the caller owns the terminal decision — `setSizeToWritten()`+transfer on success, or
  `discard()` on failure — so **every path is leak-safe**.

## Evidence
- **JMH** (`JsonBodyEncoderBenchmark`, `-prof gc`, 50-row payload): allocation per encode
  **~11432 → ~1192 B/op (~90% less)**, `gc.alloc.rate.norm` stdev ~0.4 B/op (rock-solid).
  Throughput directional-only at smoke iteration counts. The residual ~1192 B/op is
  per-response wrapper objects (LoanedBuffer handle, HttpEncodedBody) — a candidate for a
  later pooling pass, out of scope here.
- **Unit** (`JsonBodyEncoderTest`, 5): byte-equality vs `writeValueAsBytes` (small/large/
  null), the **grow** path, and **two leak paths** (serialization failure; failure *after*
  a grow) asserted via `LeakTrackingAllocator.outstanding() == 0`.
- **Lint**: `mvn -pl exeris-kernel-community pmd:check checkstyle:check` → 0/0.
- **Wall guard**: `ExerisArchitectureTest` 11/11.
- Full `http` package + all 782 community unit tests green.

> **Local caveat:** a full `mvn clean install` on this Windows/JDK-26 box crashes the
> forked VM in the native `Argon2idPasswordEncoderTckTest` (passes in isolation) — a known
> native flake unrelated to this change (no crypto touched). CI (Linux) is the authority.

## Scope
The virtual-thread-friendly Jackson recycler pool (which would trim the ~1192 B/op
residual on carrier vthreads) is deliberately **not** in this PR: per ADR-052 that is an
application concern via the existing `JsonMapperCustomizer` ServiceLoader seam, not kernel
code. This PR is purely the streaming rewrite.

## Follow-ups
- Forward-port to `development/0.11.0`.
- Version bump `0.10.1 → 0.10.2` + CHANGELOG/`docs/release` entry at release-integration
  time (kept out of this feature PR per the terse-CHANGELOG convention).
- Re-run the perfbox campaign with the fix (and a `fix-on @ 256 MB heap` variant) to
  quantify the p99-tail effect from reduced allocation independent of heap sizing.

### DIFF
--- a/docs/subsystems/http.md
+++ b/docs/subsystems/http.md
@@ -174,6 +174,21 @@ driver detail (The Wall — the seam and `ObjectMapper` never enter SPI). Scopes
 web-client scope — `KernelWebClient` is app-constructed with its own registries (ADR-034), so a
 per-instance client mapper is already an application capability.
 
+### JSON response-body zero-copy encoding (since v0.10.2)
+
+`JsonBodyEncoder.encode` streams Jackson's output **directly into the loaned off-heap response
+segment** via `SegmentSink` — an `OutputStream` over a `LoanedBuffer` that grows to a larger buffer on
+overflow — instead of serializing to a heap `byte[]` and `MemorySegment.copy`-ing it in. This removes a
+per-response heap allocation and a full extra pass over the payload; output is byte-identical and no SPI
+surface changes (an internal Community reimplementation, The Wall holds). The `java.io.OutputStream` is
+the Jackson `writeValue(OutputStream, …)` seam only: no heap buffering happens — each write lands
+straight in the off-heap segment — so this is the zero-copy path *by construction* (the bridge, not a
+buffer), not the `java.io`-on-hot-path case the guardrails warn about. Ownership is single-live-buffer:
+the sink transfers the buffer to the returned `HttpEncodedBody` on success and releases it on every
+failure path. The mirror-image `CommunityJsonRequestBodyEncoder` (client request bodies) still
+materialises-then-copies; converting it is the No-Waste-Compute follow-up, analogous to the SSE
+zero-copy-emit follow-up above.
+
 ### HTTP/2 stream admission (since v0.8 Sprint 5, HTTP-112)
 
 `Http2SessionContext.admitClientStreamId(int)` enforces two RFC 7540 invariants that the Community h2c session previously did not validate:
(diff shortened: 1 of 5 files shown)


arkstack-dev:
Thanks — addressed the two documentation/accuracy nits in ede16eb; the third is a genuine separate follow-up.

**1. `java.io.OutputStream` on a zero-copy path** — added a subsection to `docs/subsystems/http.md` recording *why* the bridge satisfies the guardrail (it's the Jackson `writeValue(OutputStream, …)` seam; no heap buffering, each write lands straight in the off-heap segment — the bridge, not a buffer), so a future reviewer sees the justification in the doc, not just the code comment. Placed next to the SSE zero-copy-emit follow-up you referenced.

**2. `INITIAL_BUFFER_BYTES = 4096`** — good catch, the comment was inaccurate: the benchmark's realistic `/users` payload already grows, so "grow is the rare exception" was wrong. Corrected the Javadoc to state that responses larger than the initial size grow transparently (one intra-buffer copy of bytes-so-far, still far cheaper than the removed heap `byte[]` + full copy), and flagged sizing/config as a tracked tuning follow-up. Left the constant itself unchanged — picking a new default without real response-size distribution would be guessing; that's exactly the follow-up.

**3. `CommunityJsonRequestBodyEncoder` mirror** — confirmed, it's the intended-out-of-scope mirror-image (client request bodies still materialise-then-copy). Now recorded explicitly in the http.md subsection as the No-Waste-Compute follow-up, so it's tracked rather than an oversight.

On verification: the lint / ArchUnit / test results in the PR body were run locally this session (`mvn -pl exeris-kernel-community pmd:check checkstyle:check` → 0/0; `ExerisArchitectureTest` 11/11; full `http` package + 782 community unit tests green). The one full-reactor caveat stands: `mvn clean install` on Windows/JDK-26 crashes the forked VM in the native `Argon2idPasswordEncoderTckTest` (passes in isolation, no crypto touched here) — CI on Linux is the authority.

PR #243, opened 2026-07-19
test(http): record Jackson writer-reuse as allocation-neutral (negative result)

arkstack-dev:
## Summary
Benchmark-only. While scoping the residual encode allocation (~1192 B/op after #241/#242), the obvious
next lever seemed to be reusing a Jackson `ObjectWriter` instead of calling `mapper.writeValue` per
request. **Measured, it does nothing.**

| strategy | `gc.alloc.rate.norm` |
|---|---|
| `STREAMING` | ~1192 B/op |
| `STREAMING_REUSED_WRITER` | ~1192 B/op |
| `MATERIALIZE_AND_COPY` (pre-#241) | ~11432 B/op |

Jackson allocates the per-call generator + `JsonWriteContext`/`IOContext`/`SerializationContextExt`
regardless of whether the call goes through the `ObjectMapper` or a reused `ObjectWriter`. Writer reuse
is **not** a lever on the residual.

This PR adds the `STREAMING_REUSED_WRITER` strategy to `JsonBodyEncoderBenchmark` and documents the
finding in-code, so nobody re-litigates writer-reuse as a production change. No production code touched.

The residual breakdown (from `-prof jfr` alloc-by-class): ~62% Jackson per-call machinery (intrinsic,
not reducible by reuse), ~35% per-response carrier/allocator objects. Of the latter, `HttpEncodedBody`
pooling is out (immutable SPI value-carrier); the allocator-side `CommunityReleaseAction` per-buffer
object is being addressed separately as a memory-subsystem change.

### DIFF
--- a/exeris-kernel-community/src/test/java/eu/exeris/kernel/community/http/JsonBodyEncoderBenchmark.java
+++ b/exeris-kernel-community/src/test/java/eu/exeris/kernel/community/http/JsonBodyEncoderBenchmark.java
@@ -26,6 +26,7 @@ import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.TearDown;
 import org.openjdk.jmh.infra.Blackhole;
 import tools.jackson.databind.ObjectMapper;
+import tools.jackson.databind.ObjectWriter;
 
 import java.lang.foreign.MemorySegment;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


arkstack-dev:
Fixed the stale doc counts in the follow-up commit: class Javadoc and the `Strategy` enum comment now list all **three** strategies, and the class doc points at `encodeStreamingReusedWriter`'s documented negative result so a reader skimming the class sees the third strategy and why it exists. Good catch — leaving "two strategies" in place would have undercut the whole point of the PR (making the finding durable). Test-only, compile + lint 0/0 green.

PR #244, opened 2026-07-19
perf(memory): fold buffer-release accounting into onRelease, drop per-buffer close-action

arkstack-dev:
## Summary
Memory-subsystem change (step 4 of the 0.10.2 residual work, but system-wide, not JSON-specific).
`CommunityMemoryAllocator` allocated a `CommunityReleaseAction` (a `Runnable`) **per buffer** and
registered it via `addCloseAction`, purely to carry the buffer's byte count plus the allocator's shared
release counters / JFR config into the release path. Its only per-instance state was `bytes` — which the
buffer already knows (`originalCapacityBytes`, the exact value `trackAllocation` adds).

Replaced with a single per-allocator `CommunityReleaseAccounting` holder (the shared counters + JFR
config); `CommunityLoanedBuffer` references it and calls `release(capacity)` from `onRelease()`. **No
per-buffer bookkeeping object, and no `addCloseAction` write+fence, on the allocation hot path** — every
allocation benefits, not just JSON encoding.

## Why it's balanced (the correctness crux)
Allocation does `trackAllocation(capacityBytes)`; release must subtract the *same* value. `CommunityArenaBuffers`
passes the requested `capacityBytes` as `originalCapacityBytes`, so `onRelease()` accounting uses the exact
value allocation added → `allocatedBytes` never drifts. Accounting and pool-return are independent, so
folding order is irrelevant.

## Behaviour preserved
- Identical `stats()` (`allocatedBytes` / `releaseCount`) and identical sampled `CommunityReleaseEvent` JFR.
- The `addCloseAction` mechanism itself is **untouched** and still available to other users (e.g. the
  leak-tracking test decorator the encoder tests rely on).
- No SPI change.

## Measured
`JsonBodyEncoderBenchmark` `STREAMING` `gc.alloc.rate.norm`: **1192.3 → 1112.1 B/op (−80 B/op**, exact GC
accounting, stdev 0.6). Honest note: this is *less* than the ~155 B/op the sampled `-prof jfr`
alloc-by-class had suggested for `CommunityReleaseAction` (sampled weight over-attributes; the GC-norm
figure is authoritative).

## Verification
- `CommunityLoanedBufferTck` (PARANOID leak, close-actions, refcount) + `CommunityMemoryAllocatorTck`
  (**MemoryStatsContract** — accounting parity) + `CommunityMemoryAllocatorTest` + encoder leak tests:
  **93 tests green**.
- `pmd:check checkstyle:check` 0/0; `ExerisArchitectureTest` 11/11.

> Windows/JDK-26 caveat unchanged: a full `mvn clean install` crashes the forked VM in native
> `Argon2idPasswordEncoderTckTest` (unrelated); memory + encoder suites verified in isolation.


arkstack-dev:
Fixed — the stray file is out of this PR's history (force-pushed `ed6dc08`; the diff is now the 5 `community/memory` files only).

You diagnosed it exactly: `SyscallLoopbackRoundTripIT.java` is a pre-existing untracked file in my working tree that got swept in by a `git add -A` I ran for a status check during commit prep — I cleaned up one stray artifact (a `jfr-out/` profiler dump) but missed this one. It's unrelated transport work, now back to untracked on disk to be handled in its own PR, not deleted.

Thanks also for verifying the accounting balance and the close-action ordering independently — and the note that `CommunityLoanedBuffer.allocateOwned(long,long,Arena)` is pre-existing dead code (`releaseAccounting = null` there is intentional for that unused path); agree it's out of scope for this PR.

PR #245, opened 2026-07-19
perf(http,memory): forward-port 0.10.2 encode/allocation batch to 0.11 (no version bump)

arkstack-dev:
## Summary
Forward-ports the 0.10.2 JSON-encode / allocator batch from `main` (0.10.x line) onto
`development/0.11.0`, mirroring the #240 pattern — **no version bump** (dev stays `0.11.0-SNAPSHOT`).
Clean cherry-picks of the four merged commits; no conflicts.

| src PR (main) | commit |
|---|---|
| #241 stream JSON **response** encoding into pooled `LoanedBuffer` | 1a87670 |
| #242 stream JSON **request** encoding into pooled `LoanedBuffer` | 0d6a247 |
| #244 fold buffer-release accounting into `onRelease`, drop per-buffer close-action | 14f7959 |
| #243 record Jackson writer-reuse as allocation-neutral (benchmark negative result) | c1cc5ae |

## What it brings
- Response + request JSON encoders stream Jackson straight into the off-heap `LoanedBuffer` via
  `SegmentSink` (no heap `byte[]`, no `MemorySegment.copy`). Byte-identical output, no SPI change.
- Allocator drops the per-buffer `CommunityReleaseAction`; release accounting folds into
  `CommunityLoanedBuffer.onRelease()` via a shared `CommunityReleaseAccounting` (−80 B/op per
  allocation, system-wide; `stats()`/JFR identical).
- `JsonBodyEncoderBenchmark` with the documented writer-reuse negative result.

## Verification on the 0.11 base
- Touched surface green: `JsonBodyEncoderTest`, `CommunityJsonRequestBodyEncoderTest`,
  `CommunityMemoryAllocatorTckTest` (incl. `MemoryStatsContract`), `CommunityLoanedBufferTckTest`
  (PARANOID leak) — 51 + memory TCK tests pass.
- `pmd:check checkstyle:check` 0/0; `ExerisArchitectureTest` 11/11.
- Version confirmed still `0.11.0-SNAPSHOT` — no bump crept in.

Each source PR was already reviewed and merged on `main`; this is a straight forward-port.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #247, opened 2026-07-20
docs(http): request encoder no longer materialises-then-copies (0.10.2 follow-up landed)

arkstack-dev:
## What

Two one-line corrections in `docs/subsystems/http.md`. Docs-only, no code.

## 1. Request encoder is no longer materialize-then-copy

The "JSON response-body zero-copy encoding (since v0.10.2)" subsection was written by #241, when only the **response** encoder had been converted, and closed with:

> The mirror-image `CommunityJsonRequestBodyEncoder` (client request bodies) still materialises-then-copies; converting it is the No-Waste-Compute follow-up…

#242 landed that conversion in the same 0.10.2 line but the paragraph was never updated, so released 0.10.2 docs describe a path that no longer exists.

**Verified against source:** `CommunityJsonRequestBodyEncoder` constructs a `SegmentSink` and calls `mapper.writeValue(sink, payload)` with the same committed-flag / `finally { sink.discard() }` ownership transfer as `JsonBodyEncoder`.

## 2. SSE zero-copy emit landed in v0.10.0

My first commit here preserved the sentence's other clause — that the SSE zero-copy-emit follow-up "remains open" — carried over from the doc rather than checked. Review caught it; verified against source, it is wrong:

- `SseEventEncoder` exposes `encodedLength(StreamEvent)` + `encodeInto(StreamEvent, MemorySegment, long)` — two passes over one shared `traverse()`/`ByteSink`, writing field-by-field straight into the egress `LoanedBuffer` segment. Both remaining `StringBuilder` occurrences in the file are javadoc prose describing the shape that was *replaced*.
- `HttpStreamEngine:166/:217` call `encodedLength`/`encodeInto` on the emit path — wired in production, not merely available. The heap `encode(StreamEvent)` wrapper has no production caller (`SseEventEncoderTest` only).
- `git log -S encodeInto` dates it to `1744622`, the v0.10.0 integration.

So the second commit drops that clause and corrects the **root** stale bullet under "Additional follow-ups" (`http.md:58`), which is where the text originated. The neighbouring **wire-framing** bullet (raw close-delimited HTTP/1.1; `Transfer-Encoding: chunked` and an HTTP/2 `DATA` path still to do) is verified still accurate and left untouched.

## Verification

Docs-only — no build, lint, or test surface. Every claim above checked against source, not against the document being corrected.

Mirrored on `development/0.11.0` in #245.

### DIFF
--- a/docs/subsystems/http.md
+++ b/docs/subsystems/http.md
@@ -55,7 +55,7 @@ Streaming (server-push) contracts — 🚧 Planned v0.10, ratified by [ADR-043](
 
 Additional follow-ups (mechanism works; refinement deferred):
 - **Wire framing** — the SSE body is written as a raw, close-delimited HTTP/1.1 response (`Connection: close`, no `Content-Length`, no chunk framing; RFC 9112 §6.3). `Transfer-Encoding: chunked` per-event framing (so SSE streams through buffering reverse proxies / CDNs that hold length-unknown responses) and an HTTP/2 `DATA`-frame path are follow-ups. The response head is HTTP/1.1-specific today.
-- **Zero-copy emit** — `SseEventEncoder.encode` currently allocates a `StringBuilder` + `String` + `byte[]` per event before the single copy into the egress `LoanedBuffer`. Encoding field-by-field directly into the `LoanedBuffer` segment (no intermediate `String`/`byte[]`) is the No-Waste-Compute follow-up for high-rate streams.
+- ~~**Zero-copy emit**~~ — **done in v0.10.0.** `SseEventEncoder` frames each event as two passes over one shared traversal: `encodedLength(StreamEvent)` measures the UTF-8 block and `encodeInto(StreamEvent, MemorySegment, long)` writes it field-by-field straight into the egress `LoanedBuffer` — no intermediate `StringBuilder`/`String`/`byte[]` and no array→segment copy. `HttpStreamEngine` calls both on the emit path. The heap `encode(StreamEvent)` wrapper is retained for tests and non-hot-path callers only.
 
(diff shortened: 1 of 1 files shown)


arkstack-dev:
Confirmed and fixed in 2c965a8 — the finding is correct on every point, and the error was mine in exactly the way that stings: I copied the SSE claim from `http.md:58` instead of checking `SseEventEncoder`, in a PR whose whole premise is "don't trust the doc, check the source" (CLAUDE.md Operating Standard #1).

Verified independently before changing anything:

| claim | check | result |
|---|---|---|
| no `StringBuilder` on the emit path | `grep -n StringBuilder SseEventEncoder.java` | 2 hits, **both javadoc prose** at :29/:31 describing the replaced shape — none in code |
| zero-copy is *wired*, not just available | `grep -n "SseEventEncoder\." HttpStreamEngine.java` | `:166 encodedLength`, `:217 encodeInto(..., buffer.segment(), 0L)` |
| heap `encode()` is non-production | `grep -rn "SseEventEncoder\." src` | only `SseEventEncoderTest` (3 call sites), zero production callers |
| predates this branch | `git log -S encodeInto` | `1744622` = v0.10.0 integration |

Changes:

1. **Dropped the SSE clause** from the 0.10.2 subsection sentence. The request-encoder statement stands on its own; no reason to graft an unrelated status claim onto it.
2. **Corrected the root bullet** at `http.md:58` in the same pass, per your suggestion — leaving it would have kept the source of the drift alive for the next person to copy. Rewritten as done-in-v0.10.0 with the `encodedLength`/`encodeInto` mechanism named, so the entry stays useful as history instead of just being deleted.

Also verified the **neighbouring wire-framing bullet** rather than assuming it: `SseEventEncoder`'s javadoc still documents the raw close-delimited HTTP/1.1 body with `Transfer-Encoding: chunked` and the HTTP/2 `DATA` path as follow-ups. Accurate, left untouched — that is the one genuinely open item in the area, as you noted.

PR description corrected too; it repeated the same false claim. Mirrored on `development/0.11.0` in #245 (`6223e8d`), verified separately against that branch's source rather than assumed identical.

PR #250, opened 2026-07-29
docs(security): KernelIsolationClaims Javadoc still mandated the fail-OPEN downgrade (v0.11 S1)

arkstack-dev:
Javadoc-only. **No bytecode change**, no behaviour change, no test change.

## What is wrong

`KernelIsolationClaims` is the normative definition of the storage-isolation claim names — it is what an
implementor reads to learn the contract. In **four** separate places it still instructed implementations to
fall back to `SHARED` on a declared-but-broken strategy:

| Location | Stale text |
|---|---|
| class-level `<h2>Fail-Closed Default</h2>` | "absent **or contains an unrecognised value** … MUST fall back to `shared(long, long)`" |
| `ISOLATION_STRATEGY` | "If absent **or unrecognised**, the Security layer MUST default to `SHARED` (fail-closed)" |
| `SCHEMA_NAME` | "If this claim is absent when `ISOLATION_STRATEGY` is `SEPARATED_SCHEMA`, the Security layer MUST fall back to `SHARED`" |
| `DATASOURCE_KEY` | same |

That is the **S-P0-07 fail-OPEN rule**, superseded by the ADR-012 §4a amendment of 2026-06-10 — and it was
never what the kernel does. `IdentityStorageMapping.fromClaims` throws `SecurityAuthenticationException`
(`EX-SEC-2002`) on all four of those inputs, and `AbstractSecurityProviderTck.IsolationStrategyContract`
pins the deny with explicit "downgrade is fail-OPEN" assertions.

So the public SPI documented a **weaker security contract than the kernel enforces**. An implementor
reading only the Javadoc would have written the downgrade the TCK rejects. ADR-012 §10's anti-drift rule
("if code differs from ADR text, update implementation plus TCK or amend ADR before merge") requires
closing this.

## What changed

The wording now separates the two cases the old text conflated:

- **absent/blank** → the one permissive fall-through (no isolation intent was expressed) → `SHARED`
- **declared but unhonourable** (unrecognised value, wrong-typed claim, missing/blank required sub-claim)
  → **terminal deny**, `EX-SEC-2002`

`StorageContextBridge` carried a second, unrelated staleness of the same class: it named
`SecurityProvider#authenticate` twice as the producer of non-`SHARED` contexts. Since ADR-040 (v0.10) the
producer is `IdentityStorageMapping.fromClaims` — verified: no `SecurityProvider` implementation reads
`KernelIsolationClaims` any more, and `fromClaims` has exactly one production caller
(`CommunityOidcIdentityProvider:129`).

## Verification

- `mvn install` on `exeris-kernel-spi` + `exeris-kernel-core` — GREEN
- `mvn pmd:check checkstyle:check` — **0 violations** (run explicitly; `verify` skips PMD)
- Javadoc rendering not verified — the repo pins no `maven-javadoc-plugin`, so it is not a build gate here.

## Relationship to the other PR

Found while writing the v0.11 S1 ADR-012 amendment (shared-scope tier), deliberately split out: this is a
security-contract correction that predates that work and deserves its own line in the history. The two PRs
touch disjoint files and can land in either order.

### DIFF
--- a/exeris-kernel-core/src/main/java/eu/exeris/kernel/core/security/StorageContextBridge.java
+++ b/exeris-kernel-core/src/main/java/eu/exeris/kernel/core/security/StorageContextBridge.java
@@ -39,18 +39,19 @@ import eu.exeris.kernel.spi.security.StorageContext;
  *
  * <p>For advanced isolation modes ({@link StorageContext.IsolationStrategy#SEPARATED_SCHEMA},
  * {@link StorageContext.IsolationStrategy#DEDICATED}), the correct
- * {@link StorageContext} is produced directly by the {@link eu.exeris.kernel.spi.security.SecurityProvider}
- * during authentication and bound by the Security Interceptor. This bridge
- * serves the common SHARED/RLS case and the tenant-less global path.
+ * {@link StorageContext} is produced from verified token claims by
+ * {@link eu.exeris.kernel.spi.security.identity.IdentityStorageMapping#fromClaims} and
+ * bound by the Security Interceptor. This bridge serves the common SHARED/RLS case
+ * and the tenant-less global path.
  *
  * <h2>SHARED-only path</h2>
  * <p>This bridge derives only SHARED contexts. For
  * {@link StorageContext.IsolationStrategy#SEPARATED_SCHEMA} and
  * {@link StorageContext.IsolationStrategy#DEDICATED} strategies the
  * {@link StorageContext} is produced exclusively by
- * {@link eu.exeris.kernel.spi.security.SecurityProvider#authenticate} and
- * bound by the Security Interceptor. Callers that require a non-SHARED
- * context MUST NOT call this bridge.
+ * {@link eu.exeris.kernel.spi.security.identity.IdentityStorageMapping#fromClaims} — the single
+ * kernel-owned fail-closed mapping site (ADR-040 §2.4, ADR-012 §4a) — and bound by the Security
+ * Interceptor. Callers that require a non-SHARED context MUST NOT call this bridge.
  *
  * <h2>JFR-First</h2>
  * <p>Every derivation emits a {@code StorageContextDerived} JFR event.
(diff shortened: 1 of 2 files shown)


arkstack-dev:
Nit verified independently and **fixed in this PR** (`dd6687c6`) — thank you, it was correct on every point and it is not really optional: it is this PR's own thesis pointed at the wording this PR added.

Confirmed against three layers:

- `VerifiedClaims#claim(String)` states it outright — *"A claim present but not representable as a single string is treated as absent — drivers that need to deny on a malformed isolation claim do so during validation, not here."*
- `CommunityOidcTokenValidator#validateIsolationStrategyWellTyped` throws `isolation-malformed` and its own Javadoc says *"Verifying it here keeps `VerifiedClaims#claim(String)` free of deny logic"*
- so a wrong-typed `ISOLATION_STRATEGY` reaches `fromClaims` as `Optional.empty()` and takes the **permissive** absent branch → `SHARED`

The class-level rule now splits the cases by enforcing layer, and `ISOLATION_STRATEGY`'s own doc no longer implies both live in one place. The Community validator is referred to **by role, not by type name**, so the SPI stays implementation-blind.

---

**Pulling the thread surfaced a bigger one, which I did not expect and which changes the severity.**

`AbstractSecurityProviderTck.IsolationStrategyContract` has six cases and **no wrong-typed case at all** — no `createTokenWithWrongTypedStrategy()` abstract factory, so no binding is ever asked for such a token. The only coverage anywhere is `CommunityIsolationStrategyResolutionTest` Case 8 (`wrongTypedStrategyClaim_denied`), a **Community-module-local unit test** that no other binding runs.

Meanwhile ADR-012 §9 has claimed since 2026-06-10 that this suite pins the deny *"on a missing sub-claim, on an unrecognized strategy value, **and on a wrong-typed claim**"*. It does not.

So the sharp edge is the exact inversion of what you'd want: the two denies the kernel enforces centrally in `fromClaims` are both pinned by the TCK, and **the one deny a driver must implement itself is the one the TCK never asks for**. A binding that omits the type-check fails open on the isolation boundary with every gate green — the "TCK anchors fail-OPEN" class from the v0.8 security audit, still open on this path.

Corrected at the normative source in #251 (`43e49dab`): §4a now names the enforcement layer per case, and §9 drops the false coverage claim and records the obligation — add the case to `IsolationStrategyContract`, demote the Community test to its binding. The fixture already exists (`TestJwt.Builder.claimRaw`, written for exactly this), so it is TCK wiring rather than new machinery. Tracking it as a separate slice; it is a test change and does not belong in a Javadoc-only PR.

`mvn install` + `pmd:check checkstyle:check` re-run after the fix — GREEN, 0 violations.

PR #254, opened 2026-07-29
feat(security): shared-scope carrier + claim + fail-closed mapping (v0.11 S2, ADR-012 §4b)

arkstack-dev:
Stream A / S2. Implements the ADR-012 amendment landed in #251. Additive throughout — no existing behaviour changes.

## Carrier

`StorageContext` gains `Optional<String> sharedScopeKey()` with a **tenant-private default**, so no existing implementor is forced to change; `ImmutableStorageContext` carries it as a 6th record component.

**Presence is the mode.** There is no visibility enum to reconcile the key against, so `mode=SHARED` with an empty key and `mode=PRIVATE` with a key present are *unrepresentable* rather than merely discouraged — §4b.2's central argument, made structural.

One new constructor invariant: **a shared scope requires an `isolationKey`**, because the write predicate pins `owner = isolationKey` and has nothing to pin to otherwise. This makes `GLOBAL` structurally incapable of carrying one.

`withSharedScope()` composes the dimension onto any existing context rather than adding a shared-scope variant of all seven factories. The axis is orthogonal to physical placement; parameterising every factory would triple the surface *and* imply the two axes are related.

## Claim

`KernelIsolationClaims.SHARED_SCOPE_KEY = "x-exeris-shared-scope"`.

## Mapping — the load-bearing part

A declared shared scope is a **terminal deny** (`EX-SEC-2002`, `shared-scope-unsupported`) in `IdentityStorageMapping.fromClaims`, because no persistence binding implements read-widen / owner-scoped-write yet.

Both alternatives are forbidden by §4b.5:

- resolving it **tenant-private** silently gives the caller less than it asked for
- **honouring** it returns a context claiming visibility that nothing enforces — the S-P0-07 class

When a binding gains the mode, this check becomes *conditional on the deployment* rather than disappearing — so there is never a window in which a declared scope resolves to anything but deny or correct enforcement.

### Documented caveat, deliberately not fixed here

A **wrong-typed** shared-scope claim collapses to absent via `VerifiedClaims.claim()` and yields tenant-private. That is a *narrowing*, not a widening, so it is safe while the enforceable answer is deny anyway — and it stops being safe the moment a binding can honour the claim. At that point type-checking it during validation becomes a driver obligation, exactly as it already is for `ISOLATION_STRATEGY` (#250, #252). Recorded in the mapping's Javadoc so S3 cannot miss it; adding the check now would gate on behaviour that does not exist yet.

## Verification

**New `IdentityStorageMappingTest` pins the mapping at SPI level — nothing did before.** It was reachable only through the Community binding, so a regression in kernel-owned logic would have surfaced as a driver failure rather than being attributed here. 7 cases: strategy fall-through and denies, shared-scope absent/declared, declared-denies-under-every-strategy (proving the orthogonal axis cannot smuggle an unenforceable scope through an otherwise-valid `SEPARATED_SCHEMA`/`DEDICATED` declaration), and blank-is-absent.

Per §9, **both** abstract suites get the contract, because §4a routes every provider through one mapping site and it must hold from both entry surfaces:

- `AbstractSecurityProviderTck.IsolationStrategyContract` — absent-stays-private + declared-denies (6 → 8 cases)
- `AbstractIdentityProviderTck` — same, in `HappyPath` and `FailurePath`

`ImmutableStorageContext` gains cases for the constructor invariant, `GLOBAL` rejection, orthogonal composition across all three strategies, and the interface default.

### Two things checked rather than assumed

1. **Blast radius of the arity change: three call sites, all in one SPI test.** The amendment claimed the migration was contained because everything constructs via factories; that is now confirmed against the compiler, not argued.
2. **`CommunitySecurityProvider` is the thin dispatcher §4a requires** — it selects an `IdentityProvider` from the registry and delegates, so the deny flows through `fromClaims` with no second mapping path. Worth confirming, since the whole amendment rests on there being exactly one mapping site.

`StorageContextBridge` now records that both its derivation paths yield an empty `sharedScopeKey` by construction — system and anonymous requests stay tenant-private with no code to audit for widening.

## Gates

- Full reactor `mvn clean install` — GREEN, nothing skipped
- `mvn pmd:check checkstyle:check` — **0 violations** (run explicitly; `verify` skips PMD)

## Note on merge order

Touches `KernelIsolationClaims` (new constant only, no edits to existing Javadoc blocks) and `AbstractSecurityProviderTck` (new case appended). **#250 and #252 touch the same two files**; conflicts should be trivial, but landing those first is the cleaner order since both are smaller and already reviewed.

Next: S3 — owner-scoped-write RLS in the Community persistence binding, which flips this deny into enforcement.


arkstack-dev:
Rebased onto `development/0.11.0` after #250 / #251 / #252 / #253 landed — force-pushed as \`28f665b3\`, now **MERGEABLE**.

Four conflicts, all additive-vs-additive (both sides appended a factory/case/switch arm): `AbstractSecurityProviderTck`, `CommunitySecurityProviderTckTest`, `CoreSecurityProviderTckTest`, `CoreSecurityProviderFixture`. Resolved by keeping both sides everywhere — nothing was dropped. Confirmed by the resulting counts: `IsolationStrategyContract` runs **9 cases** on both bindings (6 original + 1 from #252 wrong-typed + 2 from this PR). Full reactor `mvn clean install` green, `pmd:check` + `checkstyle:check` 0 violations, re-run after the rebase.

---

## Disposition of the two SonarCloud signals

Both are pre-existing repo-level conditions, not properties of this change. Neither is fixed here; both are worth their own decision.

### "1 New issue" — `java:S2187` — false positive, instance #78 of a systemic pattern

`IdentityStorageMappingTest:37`, BLOCKER, *"Add some tests to this class."* The class does contain tests — 7 of them, all inside JUnit 5 `@Nested` classes (`Strategy`, `SharedScope`), which the rule is not recognising. They run: `Tests run: 7, Failures: 0`.

This is not specific to the new file:

- **77 open `java:S2187` issues already on the main branch** (`api/issues/search?rules=java:S2187` → `total: 77`)
- **103 test classes** under the scanned `src/test` roots use the `@Nested`-only shape with zero top-level `@Test`

It surfaces as a *new* issue here only because the file is new and the rest predate the new-code period. Restructuring this one test to differ from 103 siblings would make it inconsistent with the codebase to satisfy a rule the codebase already disagrees with in 77 places, so I have not done that. The real fix is repo-level — either accept the rule's verdict once in Sonar, or scope an `sonar.issue.ignore` entry for the JUnit-5-nested shape in `sonar-project.properties`. That deletes 77 issues at once and is a policy call, not a PR call.

### "0.0% Coverage on New Code" — the coverage wiring is dead config

Not a statement about this PR's tests. `sonar-project.properties:20` does set

```
sonar.coverage.jacoco.xmlReportPaths=**/target/site/jacoco/jacoco.xml
```

and JaCoCo genuinely produces that XML (root `pom.xml`, `jacoco-report` bound to `verify`, XML+HTML formats, with per-module ratchet minimums). But **no workflow runs a Sonar scanner** — `.github/workflows/` contains `maven`, `codeql`, `dependency-review`, `claude`, `claude-code-review`, and nothing else. Analysis therefore comes from SonarCloud **Automatic Analysis**, which never builds the project, so `target/site/jacoco/jacoco.xml` does not exist when it runs and the import silently no-ops.

Consequence: coverage on new code reads 0.0% on **every** PR and always will, while the properties file reads as though coverage is tracked. Worth knowing before anyone treats that number as a signal. Closing it means adding a scanner step that runs after `verify` — a CI change, out of scope here.

Unrelated minor drift noticed in the same file: `sonar.projectVersion=0.6.0` against a reactor at `0.11.0-SNAPSHOT`.

PR #255, opened 2026-07-29
ci(sonar): run the scanner from CI so coverage is imported; suppress the @Nested S2187 noise

arkstack-dev:
Closes both repo-level Sonar conditions surfaced while dispositioning #254. Neither was a property of that PR.

> ## ⚠️ Two manual steps, or this changes nothing
>
> 1. Add a **`SONAR_TOKEN`** repository secret. Without it the step *skips* by design (so forked PRs don't fail the gate) — analysis simply stays where it is.
> 2. **Turn Automatic Analysis OFF** in the SonarQube Cloud project. It is mutually exclusive with CI analysis: leave it on and SonarQube Cloud **rejects** the CI submission, so the pipeline looks healthy while results keep coming from the old path. This is the failure mode most likely to waste an afternoon.
>
> Both are recorded in `CONTRIBUTING.md` too, since neither is expressible in-repo.

## 1. Coverage was structurally unmeasurable, not untested

`sonar-project.properties` already set `sonar.coverage.jacoco.xmlReportPaths`, and JaCoCo really does emit that XML under `-P coverage` (with per-module ratchet minimums). But **no workflow ever ran a scanner** — analysis came from SonarCloud **Automatic Analysis**, which never builds the project. So the report did not exist when the import ran, the import silently no-opped, and *Coverage on New Code* read **0.0% on every PR regardless of the tests in it**.

The properties file reading as though coverage were tracked is worse than it plainly not being configured — it invites treating a meaningless zero as a signal.

The scan now runs inside `build-and-verify`, immediately after `mvn clean verify -P coverage`, reusing that build's classes and JaCoCo XML instead of paying for a second build. Supporting changes:

- `fetch-depth: 0` on checkout — without full history Sonar cannot attribute new code or blame, and the new-code period silently degrades to "everything"
- `sonar.java.binaries` / `test.binaries` / `libraries` — the CLI scanner does not infer these. (The Maven `sonar:sonar` goal *would*, but it also ignores `sonar-project.properties` entirely, so switching to it would mean migrating all the tuned source/exclusion/CPD config into the POM — a much larger change for no gain.)
- the step is guarded on `SONAR_TOKEN` being non-empty, read from a **job-level** `env` because `secrets` is not available in a step-level `if`

## 2. `java:S2187` — 77 false BLOCKERs, suppressed **provisionally**

The rule fires on the house test shape: tests live in JUnit 5 `@Nested` inner classes, so the outer class has no top-level `@Test` and the analyzer served by Automatic Analysis does not look inside. That is **103 test classes** here and **77 open issues**, none of them real — every one of those classes runs tests.

Suppressed on `**/src/test/java/**/*.java`, with two things written into the file rather than left implicit:

- **The trade-off.** This also hides a genuinely empty test class. Judged the lesser cost against 77 false BLOCKERs, which train reviewers to ignore the rule entirely — and a truly empty test class is something review catches.
- **It may already be unnecessary.** This same PR moves analysis onto a current `sonar-java`, and newer analyzers understand `@Nested`. **Verify against the first CI-based run and delete the suppression if S2187 no longer fires.** Marked `PROVISIONAL` in the file so it does not quietly become permanent.

## Also

`sonar.projectVersion` still read `0.6.0` against a `0.11.0-SNAPSHOT` reactor.

## Verification

YAML parsed and the resulting job structure asserted (step order, the `fetch-depth: 0` on checkout, the job-level env, the guard expression). The analysis path itself can only be verified by the first run on a branch that has the secret — stated rather than claimed green.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -107,6 +107,20 @@ jobs:
         #
         # Skipped when the secret is absent (forked pull requests) — a missing token must not fail
         # the build gate.
+        #
+        # continue-on-error, deliberately: this job is the ROOT of the entire pipeline — all eight
+        # other jobs depend on it directly or transitively — so an unhealthy analysis step would take
+        # down every merge gate for a reason unrelated to the code under test. Static analysis is not
+        # a correctness gate here and must never be able to do that.
+        #
+        # It is load-bearing right now, not just prudent: the SonarQube Cloud organisation is
+        # currently without an administrator, so Automatic Analysis cannot be switched off — and
+        # SonarQube Cloud *rejects* a CI submission while it is enabled. Until that is resolved this
+        # step is expected to fail whenever a token is present.
+        #
+        # REMOVE once a CI-based analysis has been confirmed green; leaving it in permanently would
+        # let the analysis rot silently.
+        continue-on-error: true
         if: ${{ env.SONAR_TOKEN != '' }}
         uses: SonarSource/sonarqube-scan-action@v5
 
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -33,9 +33,18 @@ jobs:
       contents: read
       packages: write   # required for GitHub Packages publish
 
+    env:
+      # Job-level so the guard on the analysis step can read it — `secrets` is not available in a
+      # step-level `if`. Empty on forked PRs, which is exactly when the step must skip rather than fail.
+      SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
+        with:
+          # Sonar needs full history to attribute new code and blame; a shallow clone silently
+          # degrades the new-code period into "everything".
+          fetch-depth: 0
 
       - name: Cache JDK tarball
         id: jdk-cache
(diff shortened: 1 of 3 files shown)


arkstack-dev:
Blocker confirmed and fixed in `2f6b44de`, along with the two non-blocking notes. One of them I checked and it does not hold — evidence below.

## Blocker — `sonar.organization`: correct, and the risk was understated

You were right that nothing sets it and that Automatic Analysis resolved it server-side. Added `sonar.organization=exeris-systems`, key read from the SonarCloud API rather than guessed.

Your dependency argument was directionally right but **too mild**. `build-and-verify` is not depended on by a couple of gates — it is the **root of the entire graph**:

```
build-and-verify ──┬─ persistence-rls-gate ─ kafka-integration-gate ─ recovery-continuity-gate ─ transport-stress-gate
                   ├─ tls-openssl-matrix
                   ├─ benchmarks ──┐
                   └─ parse-jfr-to-json ─ publish-gh-pages-data
```

All eight other jobs hang off it. A failing analysis step takes down the whole pipeline over something that is not a correctness gate.

**So I took both of your options, not either/or** — and `continue-on-error: true` is load-bearing here rather than a safety net, for a reason that landed after this PR was opened:

> The SonarQube Cloud **organisation currently has no administrator** — the previous admin account stopped being a member of the organisation. A support ticket is open. Automatic Analysis therefore **cannot be switched off**, and Cloud rejects a CI submission while it is enabled.

So adding `SONAR_TOKEN` today would fail the step on the first run on a protected branch — exactly the scenario you predicted, now with a known cause rather than a suspicion. The flag is commented `REMOVE once a CI-based analysis has been confirmed green`, so it does not become permanent cover for a rotting analysis.

`CONTRIBUTING.md` now **orders** the two manual steps instead of listing them: disabling Automatic Analysis is a *precondition* of adding the token, not a companion task. Until the org is fixed, no token is the state you want — the step skips rather than fails.

## Non-blocking 1 — `projectVersion`: taken

Matched exactly: `0.11.0-SNAPSHOT`.

The drift class you point at is real and matching the qualifier only resets it. The durable fix is injecting the version from the POM at scan time, which I deliberately did **not** add now — putting another moving part into CI is hard to justify while the analysis path itself cannot be exercised. Worth doing once the org is unblocked.

## Non-blocking 2 — the "77 BLOCKERs" claim: checked, and it stands

Fair thing to question, since S2187 is `Major` in Sonar's default Java profile. But the number came from the API, not from the default profile:

```
GET /api/issues/search?componentKeys=exeris-systems_exeris-kernel&rules=java:S2187&facets=severities

total: 77
severity distribution: {'BLOCKER': 77}
impacts: [{'softwareQuality': 'MAINTAINABILITY', 'severity': 'BLOCKER'}]
```

All 77 report as BLOCKER on this project, so the comment stands as written. Leaving the query in this thread so the next person can re-check it in one call instead of re-deriving it.

## `fetch-depth: 0`

Agreed on the trade — one-time clone cost against correct new-code and blame attribution. Worth it, and it is the difference between a meaningful new-code period and one that silently degrades to "everything".

PR #256, opened 2026-07-29
feat(persistence): publish exeris.shared_scope so RLS can widen reads while writes stay pinned (v0.11 S3, ADR-012 §4b.4)

arkstack-dev:
Stream A's last mechanism slice. Community persistence side of the shared-scope tier.

## The shape turned out simpler than the plan assumed

The kernel ships **no RLS policies** — the deployment owns them, and tenant isolation already works by publishing `exeris.tenant_id` for the policy to read. Shared scope needs nothing more exotic: publish a second session setting and let the policy widen on it.

The useful discovery is that **`WITH CHECK` already expresses owner-scoped write**. A conforming policy widens `USING` and leaves `WITH CHECK` byte-for-byte identical to the tenant-private one:

```sql
CREATE POLICY shared_scope_isolation ON <table>
  USING (tenant_id = current_setting('exeris.tenant_id', true)
         OR (NULLIF(current_setting('exeris.shared_scope', true), '') IS NOT NULL
             AND shared_scope = current_setting('exeris.shared_scope', true)))
  WITH CHECK (tenant_id = current_setting('exeris.tenant_id', true));
```

So "read widens, write pins" needs no new write machinery at all — widening reads does not require relaxing writes. The policy is documented on the interceptor as normative and exercised by the new IT.

## Published on every strategy — and unconditionally

Visibility is orthogonal to placement, so the setting rides all three strategies. SHARED carries it in the statement that already existed (still one round-trip); SEPARATED_SCHEMA and DEDICATED each pay one more. **DEDICATED previously issued no SQL at all**, so that is a genuine cost and the class Javadoc no longer claims otherwise.

**The unconditional part is the security-critical one, not a detail.** `set_config(..., false)` is *session*-scoped, so on a pooled connection a previous request's shared scope survives into the next. Skipping the statement when no scope is declared would silently widen a request that never asked to participate — the same pool-recycle hazard the tenant key already guards against by always writing a value. Absence is published as `""`, and both the unit tests and the IT pin that specific case.

## Verification

Unit tests (9 green): publish, **clear-when-absent**, and per-strategy routing.

`CommunityPersistenceSharedScopeIT` covers the §4b.4 access matrix against a live Postgres — read-widen, write-pin (a tenant *inside* the partition still cannot forge a partition-mate's row), no-pool-bleed, and tenant-private unchanged. Cross-tenant mutation is asserted as **denied, not supported**: out of contract scope per §4b.4, and the assertion exists so it cannot creep in later.

> **The IT compiles but has not run here.** Docker is unavailable in this environment, so it reports as *skipped*, not passed. `persistence-rls-gate` already runs every `integration`-tagged test in this module, so it picks the IT up with no workflow change — **that gate is the actual proof**, not this description.

Full reactor `mvn clean install` green; `pmd:check` + `checkstyle:check` 0 violations.

## PMD pushed back, and it was right

The class total hit 33 against a 30 ceiling once these paths landed. Rather than raise the ceiling, PostgreSQL identifier validation moved into its own `PostgresIdentifier` class — it guards `SET search_path` interpolation, which cannot use a bind parameter, and was never an interception concern. The interceptor also stopped duplicating the prepare/bind/execute/translate skeleton between the two settings.

## Deliberately NOT in this PR

`IdentityStorageMapping` still denies every declared shared scope, so there is no end-to-end path yet — by design.

Flipping that deny needs a signal for *"this deployment's policy honours shared scope"*, and **the kernel cannot introspect it**: the policy lives in the application's DDL, not in anything the kernel ships or reads. Answering it means a cross-subsystem capability seam between Security and Persistence, which deserves its own decision rather than being improvised at the end of this slice. Until then the fail-closed posture from #254 stands, which is the safe state to sit in.

That leaves Stream A's mechanism complete on both sides — carrier and enforcement — with one explicitly-scoped decision between them and the tier going live.

### DIFF
--- a/docs/adr/ADR-012-security-trust-model-upgrade-for-resource-server-validation-and-fail-closed-runtime.md
+++ b/docs/adr/ADR-012-security-trust-model-upgrade-for-resource-server-validation-and-fail-closed-runtime.md
@@ -165,10 +165,18 @@ the RFC deferred to this amendment. **Not yet implemented** — see §10.
   through one mapping site — the contract must be proven from both entry surfaces.
 - `ImmutableStorageContext` must have a constructor-invariant case: `sharedScopeKey` present with
   `isolationKey` absent is rejected (§4b.2).
-- `AbstractPersistenceEngineTck` must carry a shared-vs-tenant **access matrix**: the read-widen path (a
+- The persistence TCK must carry a shared-vs-tenant **access matrix**: the read-widen path (a
   tenant reads another owner's row inside the same shared scope) and the write-pin path (a tenant cannot
   write a row owned by another tenant, inside or outside the shared scope). Cross-tenant mutation stays out
   of scope (§4b.4) and MUST NOT be added to the matrix as an allowed cell.
+  - **Host corrected 2026-07-29 (v0.11 S3): `AbstractSharedScopeAccessMatrixTck`, not
+    `AbstractPersistenceEngineTck`.** This clause originally named the latter. Implementation showed that
+    would be self-defeating: the only in-repo binding of `AbstractPersistenceEngineTck` runs on H2 in
+    PostgreSQL-compatibility mode, and H2 implements neither `CREATE POLICY` nor `current_setting`, so a
+    matrix hosted there could only ever *skip*. That is a vacuous anchor — the same failure class as the
+    wrong-typed isolation case, which sat unbound behind a green suite until 2026-07-29. The matrix
+    therefore lives in its own abstract suite whose binding is a live-database one, which keeps the
+    obligation both abstract and non-vacuous. The requirement is unchanged; only its host is.
 - **"Two bindings" — RULED (amended 2026-07-29).** The §9 two-binding language means *every binding that
   exists in-repo, plus a recorded obligation on out-of-repo bindings* — not "block the contract until a
   second in-repo binding is invented". Only one in-repo persistence binding exists (Community/Postgres), so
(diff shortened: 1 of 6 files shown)


arkstack-dev:
Blocker accepted and fixed in `31a6dcc0` — with one deliberate deviation from the letter of §9, recorded in the ADR rather than glossed over.

You were right, and it stings appropriately: §9 was written the same day by the same hand, and this PR shipped the matrix as a Community-local IT. A binding-local test proves nothing about an out-of-repo binding.

## The matrix is now abstract

`AbstractSharedScopeAccessMatrixTck` — read-widen, write-pin, no-bleed, tenant-private-unchanged, with cross-tenant mutation present only as a denial a binding MUST NOT relax. `CommunityPersistenceSharedScopeIT` becomes its live-PostgreSQL binding, owning only the table, the conforming policy, and the two store operations.

## Why not `AbstractPersistenceEngineTck`, which §9 named

Because hosting it there would have been **self-defeating**, and I'd rather say so than quietly comply.

That suite's only in-repo binding — `CommunityPersistenceEngineTckTest` — runs on **H2 in PostgreSQL-compatibility mode**. H2 implements neither `CREATE POLICY` nor `current_setting`. A matrix hosted there could only ever `assumeTrue(...)`-skip.

That is a **vacuous anchor** — precisely the failure class this repo fixed *earlier today* in #252, where the wrong-typed isolation deny sat unbound behind a green suite because the abstract suite never asked any binding for that token. Satisfying §9 literally would have re-created the exact bug the same milestone just closed.

A dedicated suite with a live-database binding keeps the obligation **abstract and non-vacuous at the same time**, which is what §9 was actually for. The requirement is unchanged; only its host is — and §9 now records the correction and the reason, so the next reader doesn't re-derive it or "fix" it back.

## §10 staleness — also closed

You were right that it was about to become false on merge. Split into:

- **implemented** — the enforcement substrate: Community publishes the setting, `WITH CHECK` pins writes, the matrix is bound
- **not implemented** — and here the original text named the wrong thing. The gap isn't "no binding enforces"; enforcement now exists. It is that **the kernel ships no RLS policy and cannot introspect the deployment's**, so it cannot tell whether a shared scope is honoured. That missing capability seam is the real reason `IdentityStorageMapping` still denies, and naming it correctly is what keeps the next slice pointed at the right problem.

## Nits

- `executeSetConfig`'s `diagnosticKey`-before-varargs shape: fair point on mis-ordering risk. Left as-is — both call sites are covered and inverting the order would put the vararg in the middle, which Java doesn't allow. Worth a small carrier type if a third caller appears.
- The IT's scenarios are no longer duplicated logic — they *are* the abstract suite's, which is the outcome you asked for.

Also moved seeding to `@BeforeEach` with a truncate: the write-pin case deliberately attempts a refused write, so the cases must not inherit each other's residue.

**Gates:** full reactor `mvn clean install` green, `pmd:check` + `checkstyle:check` 0 violations. The matrix still cannot execute here — no Docker, 4 skipped — so `persistence-rls-gate` remains the proof, not this comment.

PR #257, opened 2026-07-29
feat(security): rule the shared-scope enforceability signal as an operator declaration (ADR-012 §4b.7)

arkstack-dev:
Closes the gap #256 deliberately left open. §4b.5 requires denying a declared shared scope wherever it cannot be enforced — which presupposes knowing whether it can be. Nothing answered that, so the deny was unconditional and the tier was unreachable end-to-end.

## The kernel cannot know, so it must not pretend to

It ships no RLS policy and cannot introspect the one a deployment wrote — the read-widen / write-pin predicates live in the application's DDL, which the kernel never sees. The signal therefore comes from the only party that *does* know: the deployment asserts it via `exeris.security.shared-scope.enforced`.

Absent or `false` denies, so every deployment that has not opted in stays fail-closed by default. The flag is fixed at provider construction rather than mutable — it takes part in a per-request security decision, and reconfigurable security state behind a live provider is its own bug.

## Two more plausible designs, rejected — and why is in the ADR

**A capability on the persistence provider** (`PersistenceEngine.supportsSharedScope()` resolved at bootstrap) reads as cleaner and more kernel-owned. It is worse. The engine does not write the policy either, so it cannot answer truthfully; its value would have to come from configuration anyway, and returning it through an engine method would **dress an operator's claim as a kernel guarantee**. That is the same category of error as documenting a contract the code does not enforce — which this milestone has already had to fix twice (#250, #252).

**Probing `pg_policies`** is the only option that genuinely verifies, and fails on every other axis: PostgreSQL-specific, defeated by a policy present on some tables and not others, needs a live database during bootstrap, and puts DDL introspection inside the kernel.

## A consequence that has now gone live

The wrong-typed shared-scope caveat from #254 said it was tolerable *"while the enforceable answer is deny anyway"* — collapsing to absent was merely a narrowing. **That condition no longer holds.** In an enforcing deployment, a caller whose scope claim is malformed silently loses the visibility it asked for instead of being told.

Type-checking that claim during token validation is therefore now a `TokenValidator` obligation on the same footing as `ISOLATION_STRATEGY` — the mapping structurally cannot make it, because `VerifiedClaims.claim()` reports a wrong-typed claim as absent. Recorded in §4b.7 and in the mapping's Javadoc so it cannot lapse. Wiring that check into the Community validator (and its TCK case, mirroring #252) is the natural follow-up.

## Compatibility

The three-argument `fromClaims` stays and delegates with `false`, so an un-migrated driver cannot silently begin honouring shared scopes. Pinned by its own test rather than left to inspection.

## Verification

`IdentityStorageMappingTest` — 11 green, including the enforced path, composition with a physical strategy (the axes stay orthogonal under enforcement too), that enforcement weakens **no other** fail-closed rule (an unrecognised strategy still denies — the flag must not become a blanket relaxation), and that the legacy overload still denies.

Full reactor `mvn clean install` green; `pmd:check` + `checkstyle:check` 0 violations.

## Note on merge order

ADR edits are confined to a **new §4b.7** on purpose. #256 rewrites §10 and is approved awaiting gates, so §10's final wording is left to whichever of the two lands second rather than manufacturing a conflict in an approved PR. Whoever merges second gets a one-line touch-up: with this seam in, the reason `IdentityStorageMapping` denies is no longer "no signal exists" but "this deployment has not asserted one".


arkstack-dev:
**#256 merged while this was open**, so the §10 touch-up its description predicted has landed here instead (`a7519219`, rebased onto `development/0.11.0`). Rebase was clean — §4b.7 is a new section, and I had deliberately kept out of §10 for exactly this reason.

Two ADR statements were falsified by this PR and are now corrected:

- **§10** said the missing piece was *"a signal for whether a deployment honours a shared scope … what `IdentityStorageMapping` still lacks"*. True when #256 wrote it; false now. Replaced with what is genuinely still open, which is narrower and sharper — the driver-side type check for a wrong-typed shared-scope claim.
- **§4b header** still carried *"Not yet implemented — see §10"* from #251. It has been stale since #254 and was now pointing at a section that contradicts it.

Full reactor green after the rebase.

**Stream A is complete end-to-end with this**: carrier (#254), RLS enforcement (#256), and the signal connecting them (#257) — with the fail-closed default intact, so no existing deployment moves off tenant-private without opting in.

The one thing I would not call finished is the wrong-typed type check, now recorded in §10 as an open obligation rather than left in a PR comment. It is a small, well-shaped follow-up: the validator change plus a TCK case, mirroring exactly what #252 did for `ISOLATION_STRATEGY`.

arkstack-dev:
Both review nits addressed in `04406645`.

**1. Enforced path had no executable coverage — fixed, and it was hiding something.**

`AbstractIdentityProviderTck` gains `createSharedScopeEnforcingProvider()` plus three cases: a declared scope reaches the resolved `StorageContext`, enforcement invents no scope for a token declaring none, and enforcement weakens no validation gate (an enforcing instance that rebuilt its validator would relax the gates guarding the tier it just widened).

The factory is **abstract, not a hook defaulting to skip**. §4a routes every provider through the one kernel-owned mapping, so no binding is exempt from the enforced path, and a skipping default would be vacuous for exactly the bindings that never looked at it — the #252 failure class.

Both sides of the seam now run off the same token buffer, so a binding that ignores its flag cannot satisfy both. Confirmed non-vacuous by mutation — pinning the 4th argument to `false`:

```
Tests run: 11, Failures: 0, Errors: 1
declaredSharedScopeIsCarried -- ERROR!
SecurityAuthenticationException: Token validation failed
  at IdentityStorageMapping.resolveSharedScope(IdentityStorageMapping.java:184)
```

One test, failing inside `resolveSharedScope` — the deny path, which is the regression itself.

**Related finding.** Chasing that case surfaced two rationales this PR had falsified: both TCKs anchored the shared-scope deny on *"no persistence binding implements read-widen / owner-scoped-write yet"*, true until #256 shipped the mechanism. The deny now rests on this provider not being constructed as enforcing. `AbstractSecurityProviderTck` had **predicted** the inversion (*"it does not disappear, it becomes conditional on the deployment"*) — it now records that it happened rather than still awaiting it.

**2. Config key documented as if wired — fixed.**

Both the Javadoc and ADR §4b.7 now state the key is named but unread: the assertion reaches the mapping through explicit construction, and binding it to configuration lands with the same wiring step that still owns issuer / audience / JWKS endpoint. Setting the property today does nothing, and the docs say so. The constant still lands now so the name an operator sets and the name that step reads are fixed together.

**Verification:** full reactor `mvn clean install` green with lint gates active (no skip flags); `ExerisArchitectureTest` 11/11; `AbstractIdentityProviderTck` bindings green incl. `SharedScopeEnforcement` 3/3; `IsolationStrategyContract` 9/9; `IdentityStorageMappingTest` 11/11.

Untouched: the wrong-typed `TokenValidator` obligation stays the recorded follow-up in §10 — it needs a validator change, not a test, so it does not belong in this PR.

PR #260, opened 2026-07-30
docs(adr): ADR-057 — JobScheduler SPI on virtual threads, no scheduled executor (v0.11 stream C)

arkstack-dev:
Opens Tier-1 stream C with its docs-only slice. Number reserved first in [exeris-docs#38](https://github.com/exeris-systems/exeris-docs/pull/38), content second.

Marked **ACCEPTED** on the same single-decider accepted-on-merge pattern as ADR-056. Two of its rulings **deviate from the ROADMAP entry**, so if either reads wrong to you, that's the thing to push back on — the rest follows from them.

## Deviation 1 — dispatch on virtual threads, not `StructuredTaskScope`

The ROADMAP sketches the Community driver on STS. STS is the kernel's only remaining preview dependency, and the default line carries exactly **four** import sites: `CommunityEventLoop`, `SubsystemOrchestrator`, `OutboxOrchestrator`, `InMemoryEventBus`. (Counted by import, not by mention — there are 18 files that *name* it, most in Javadoc.) A new subsystem built on STS would be a fifth, inside the milestone whose stated direction is removing them.

The ROADMAP already applies exactly this reasoning to the anticipated `KafkaEventLoop` — *"build it on the GA structured-concurrency layer from the start"*. It applies with more force here, because nothing is written yet. An ArchUnit rule pins it instead of leaving it to reviewer memory.

## Deviation 2 — no `ScheduledExecutorService`, so no scoped-ban exception

The ROADMAP asks for the STPE carve-out **and** for an injectable time source. Those are mutually exclusive, and that is the whole argument: `ScheduledThreadPoolExecutor` computes its deadlines from `System.nanoTime()` internally, with no seam to displace it. A driver timing through STPE **cannot** be driven by an injected clock — which puts the entry's own merge gate, a deterministic trigger TCK with no `Thread.sleep`, out of reach through it.

So the driver owns its timing loop: a delay-ordered queue and one dispatcher virtual thread. **Removing the exception beats documenting it** — the scoped-ban list is worth more while it stays empty.

Obligation 4 goes one step further than the ROADMAP: a time source alone isn't enough for determinism, because a test that advances a clock while the dispatcher sleeps on a real monitor still waits in wall-clock time. **The wait primitive is injected too.**

## Also ruled

- **Identity crosses the boundary by capture at submission** — and a job dispatched with *no* captured context **fails closed** rather than running under an ambient identity. That is ADR-012 §4a's reasoning applied to dispatch: a resolution the kernel cannot honour is a deny, not a downgrade. The ADR also says plainly that the captured context is a **snapshot**, so a job firing an hour later may run under authority since revoked — recorded rather than left to be discovered.
- **`JobHandle.cancel` + shutdown drain are the structured lifecycle boundary**, satisfying CLAUDE.md's unstructured-concurrency rule without depending on any orchestration helper.
- **Cron is the standard five fields, full stop** — no seconds, no `@reboot`, no extensions.

## Found while writing: a stale guard rationale

`ExerisArchitectureTest.noExecutorsAnywhere` states its reason as *"All concurrency must use `StructuredTaskScope` (JEP 525)"*. That predates the Platform Baseline and now contradicts it. **The rule stays — only its stated reason is stale**, and correcting the text rides the SPI PR. Flagging it here because a rule whose rationale argues for the thing the milestone is removing is the kind of guard someone eventually "fixes" in the wrong direction.

## Trade-offs, recorded not argued away

The driver now owns timing code the JDK already ships correctly — drift on periodic triggers, clock adjustment, and cancellation races are exactly the corners STPE gets right, and the TCK has to carry that cost. And bare virtual threads give containment **by rule** where a structured scope gives it **by construction**.

## Verification

Docs-only. Cross-references checked to resolve: ADR-012, ADR-056, `docs/subsystems/flow.md`, ROADMAP §"Platform Baseline for 1.0 GA" (line 1955), CLAUDE.md §"Scoped Bans". The four STS sites and the `noExecutorsAnywhere` rule text were read from source, not from the ROADMAP. The JFR single-phase claim cites the existing in-code precedent in `CommunityOidcIdentityProvider.authenticate`.

Next slice (S8) is the SPI + Loom driver + `AbstractJobSchedulerTck`.


arkstack-dev:
Nit closed in `c3536007` — the finding was correct, and it was an oversight, not a considered cut.

**Excluded rather than restored**, and the reasoning lands on a slightly different axis than the one suggested: **coupling direction**.

"Run this job when event X occurs" is already expressible — an event handler calls `submit`, and the job gains context capture, a handle, cancellation, and JFR events by that route. A trigger *kind* would instead make `spi.scheduling` depend on `spi.events` to name the subscription: an events dependency inside the scheduling contract, bought for something composition already covers. Events submitting jobs keeps both contracts narrow; the reverse is a seam the kernel maintains forever for no capability gained.

Also named the part of that request that *is* genuinely scheduler-shaped — debounce and rate-limit semantics — as **dispatch policy rather than a trigger kind**, and out of v0.11 either way, so it doesn't come back as "well, event-driven meant *this*".

Placed next to the Flow/DAG exclusion so both boundary arguments read together: the scheduler composes with its neighbours instead of absorbing them. Obligation 7 now states the trigger set explicitly and points at the fence, and the ROADMAP annotation carries it as a third departure — so the trigger list can't drift before S8 encodes `JobTrigger` off this ADR.

Thanks for checking the STS count, the guard rationale, and the Platform-Baseline scoping against source rather than against the PR body — particularly the observation that the GA fork/join layer is scoped to the four existing sites and is *not* offered as a primitive for new subsystems. That's the reading Deviation 1 depends on, and it's worth having stated by someone other than the author.

PR #261, opened 2026-07-30
feat(storage): BlobStorageProvider SPI + filesystem driver + TCK (v0.11 S5, ADR-056)

arkstack-dev:
Second Stream B slice — the SPI ADR-056 ruled, its first driver, and the executable contract. Opens `eu.exeris.kernel.spi.storage.blob`, the 17th top-level SPI package.

## Two ADR obligations came out differently once implemented

Both are **amended in the ADR in this PR** rather than left describing something the code doesn't do.

**Obligation 3 specified a `retain()`/`close()` protocol per direction. It turned out no hand-off is needed at all.** `BlobUploadHandle.write` takes a `MemorySegment` and must not retain it; `BlobDownloadHandle.read` fills a caller-supplied one. Both mirror `TransportStream.write`/`read` exactly, including the non-positive no-op and `-1` at EOF, so a reader who knows the transport seam knows this one.

Why it's better and not just simpler: one pooled `LoanedBuffer` drives an entire transfer (per-chunk ownership transfer would force a fresh allocation per chunk), and with nothing crossing the seam owning memory **there is no `retain()` to forget**. The trade-off the ADR called *"the SPI's sharpest edge"* is **retired, not mitigated**. The PARANOID TCK stays and now proves the absence.

**Obligation 6 made key-injection defence a driver duty. It's discharged one level earlier.** `BlobRef` rejects traversal, absolute keys, separators, empty segments and NUL at construction, so an unsafe reference is *unrepresentable* rather than merely forbidden — no driver can forget a check it cannot reach. The FS driver still verifies containment after resolution, as a backstop that holds if the carrier is ever relaxed.

## Isolation is structural

`BlobRef` carries no tenant; it resolves against the ambient `StorageContext` **inside** the store. A forged or replayed reference lands in the resolving caller's own namespace — there is no cross-tenant reference to reject because there is no way to write one down.

An absent isolation key **denies** (`EX-BLOB-8002`) rather than falling back to an unscoped root. `ImmutableStorageContext.GLOBAL` leaves it empty, and that fallback would place one tenant's object where every tenant can reach it — reached by silence rather than by a decision.

## Both new guards proven non-vacuous by mutation

| Mutation | Result |
|---|---|
| Resolve every tenant to one fixed directory | **exactly 3** failures, all in `Isolation`; other 15 green |
| Add a `java.nio.file` reference to the storage SPI | `noFilesystemTypesInStorageSpi` fails |

The arch rule bans **`java.nio.file` only** — `noJavaIoInSpi` already covers `java.io` repo-wide, so restating it would be duplicated guard noise — and is scoped to the storage package because `CryptoProviderConfig` legitimately carries a `Path`.

## Notes

- Filesystem **placement** lives in its own collaborator, not the store. Tenant encoding, the containment check and the sidecar are the security-relevant half; separating them means a change to placement is not read as a change to I/O. (It also cleared a real `TooManyMethods`/complexity finding — the split was the fix, not a suppression.)
- Tenant directories are **hex-encoded**, because an isolation key of `..` would otherwise be a directory name that escapes the root. Encoding removes the question instead of enumerating bad values.
- Uploads stage to `.uploading` and atomically move on commit, so an interrupted transfer leaves nothing visible.
- `docs/subsystems/storage.md` lands here, per the one-contract-doc-per-subsystem convention the ADR's protocol required.

## Verification

Full reactor `mvn clean install` green **with lint gates active** (no skip flags); `ExerisArchitectureTest` **12/12**; `AbstractBlobStorageTck` **18/18** on the FS binding under `LeakDetectionMode.PARANOID`; `BlobRefTest` **19/19**.

One process note worth recording: an earlier "green" run of mine executed in the main worktree rather than this branch's, so it proved nothing — caught by `ExerisArchitectureTest` reporting 11 rather than 12. Every count above is from a build re-run with an explicit path into this worktree.

Next slice (S6) is the S3-compatible driver that closes the ≥2-binding gate.


arkstack-dev:
Both blockers closed, plus the three nits. Pushed as `0ad83621`.

## 1. JFR failure telemetry

The finding was right and the precedent you cited is the one I followed. Fourteen throw sites emitted nothing.

`CommunityBlobFailures` is now the single choke point — each factory **emits the event and returns the exception**, so a site reads `throw CommunityBlobFailures.transferFailed(...)`. This deviates from `CommunityIdentityJfrEvents`, which emits *beside* the throw. At fourteen sites that shape makes "recorded but not thrown" and "thrown but not recorded" both reachable by omission; returning the exception removes the pairing rather than documenting it. Both events are single-phase commits — transfers block on a channel on virtual threads, so a `begin() → I/O → commit()` straddle would bind the carrier-local `EventWriter` across a park.

Writing it surfaced a **live defect in the deny reason**. `ImmutableStorageContext.GLOBAL` carries strategy `SHARED` with an empty key, so `isolationDenied(provider, context.strategy().name())` reported `SHARED` for exactly the case where the point is that *nothing was scoped*. `reason` is now a classification (`no-isolation-key` / `path-escape`) and the strategy is recorded beside it. This also improves `EX-BLOB-8002` rawArgs, which carried the same misleading value.

Registered in `telemetry.md` §Required Events and in a new `storage.md` §Telemetry. ADR-056's Engineering Protocol gains the item it was missing, with a note on why it landed here rather than in S6 as the milestone plan scheduled: the failure sites ship *now*, and landing blind failure paths to instrument them a slice later is the worse order.

## 2. Anti-traversal test — and it falsified my own rationale

`CommunityFilesystemBlobLayoutTest` binds hostile isolation keys directly. **The mutation proof did not say what I expected:**

| Mutation | Result |
|---|---|
| Strip hex encoding (`encoded = isolationKey`) | **3 of 7 fail** — deep-traversal, separator-bearing, injectivity |
| … the bare `".."` case | **still passes** |

The `t-` prefix already makes `t-..` an ordinary directory name, so the bare `..` never was a traversal. The claim in `storage.md` and the layout Javadoc — *"a value of `..` would otherwise be a directory name that escapes the root"* — was **wrong**, and it was the sentence the PR description leaned on hardest. Both are rewritten around what actually discriminates: a key carrying separators (`../../etc`, or merely `a/b`) becoming a nested chain, and the mapping staying injective. `storage.md` now records the mutation result inline so the claim carries its own falsification test.

Removing the two emits fails both JFR cases on `No value present`.

## 3. Nits

- **EP drift** — fixed, with the reasoning: once obligation 6 moved into `BlobRef`'s constructor, a driver-level test would have had to construct an invalid carrier to exercise it, which the carrier forbids. `BlobRefTest` holds it.
- **Obligation 2 overstates wiring** — restated in the future tense. Your read of the `GraphProvider` precedent is right and the ADR now names it, along with where the slots land (with the second driver, where provider selection first has something to select between).
- **Lone `.` segment** — now rejected, for the reason you gave: unspecified, and it means two different things depending on whether a driver resolves the key or stores it verbatim. Also rejected as a container.

## Verification

Full reactor `mvn clean install` green with lint gates active (no skip flags). `ExerisArchitectureTest` 12/12. `AbstractBlobStorageTck` 18 under `PARANOID`. `BlobRefTest` 22 (was 19). Layout 7 and JFR 2, both new.

PR #262, opened 2026-07-31
feat(scheduling): JobScheduler SPI + Loom driver + TCK (v0.11 S8, ADR-057)

arkstack-dev:
Stream C, slice S8. Implements ADR-057: the scheduling SPI, the Community Loom driver, and `AbstractJobSchedulerTck` with its binding.

## What landed

`eu.exeris.kernel.spi.scheduling` — `JobSchedulerProvider` discovers, `JobScheduler` operates, plus `JobDescriptor` / `JobHandle` / `JobState` / `JobTrigger` and `EX-JOB-9001..9003`. Community driver dispatches in-process on virtual threads. Contract doc at `docs/subsystems/scheduling.md`.

**No `ScheduledExecutorService`, no `StructuredTaskScope`** (ADR-057 §2-3). The driver owns a delay-ordered queue and one dispatcher virtual thread against an injected clock.

The seam injects **the wait primitive as well as the time source**. A clock alone would not have been enough: a test that advances time while the dispatcher sleeps on a real monitor still waits in wall-clock time. `advanceTime` is abstract on the TCK rather than defaulted-to-sleep, so a binding cannot pass the suite while sitting on real time — which is the coverage the whole design exists to buy.

## Two real bugs the tests found

Both were caught by contract cases, not by review:

1. **Cancellation silently did nothing on repeating jobs.** A repeating job was requeued in state `COMPLETED`; `cancel()` read that as terminal, refused, and left the job in the queue — so it kept firing. It now returns to `SCHEDULED`, with `FAILED` preserved so the last outcome stays observable.
2. **Overlapping runs.** Rescheduling at dispatch time let a body that overran its interval start again before the first run returned — fixed-*rate* behaviour, which `FixedInterval` explicitly documents it does not offer. A job now leaves the queue for the whole run and returns only when it finishes, so it is either queued or running and never both.

## A mutation that exposed my own test as vacuous

Disabling the §5 fail-closed guard **initially still passed**. With the guard gone, `rebindAndRun` dereferenced a null carrier, so the body did not run — the test passed for the wrong reason and would not have caught a real regression.

The unrebound path now runs the body instead of NPE-ing, which is exactly what §5 forbids, so removing the guard is observable as the contract violation it is. All three mutations now kill their intended case:

| Mutation | Result |
|---|---|
| Disable the §5 fail-closed guard | `contextlessJobFailsClosed` fails |
| Intersect day-of-month with day-of-week | `domAndDowAreUnioned` fails |
| Reference `StructuredTaskScope` in the driver | `noStructuredTaskScope` fails |

## `ExerisArchitectureTest` cannot enforce anything about Core or Community

This is wider than scheduling and I want it visible rather than buried.

The suite runs in `exeris-kernel-tck`, whose only production dependency is the SPI. No Core or Community class is ever on its analysis classpath, so a rule naming one **can never fire** — my first attempt at the ADR-057 §2 guard sat there passing on an empty set.

`noThreadLocal`, `noExecutorsAnywhere`, `noCompletableFuture` and `noUnsafe` are all scoped `eu.exeris.kernel..` and read as repo-wide. They cover SPI and TCK only. `CLAUDE.md` §"Scoped Bans" states the `ThreadLocal` and `Executors` bans as ArchUnit-enforced.

I did **not** fix that here — it needs ArchUnit wired into Core and Community and will surface pre-existing violations, which is its own slice. It is recorded in the ROADMAP alongside a second finding: `noExecutorsAnywhere`'s stated reason ("All concurrency must use `StructuredTaskScope`") now contradicts the Platform Baseline directly above it.

The driver guards therefore live in `CommunitySchedulingArchitectureTest`, in the module that can see them, and assert a non-empty analysis set before asserting anything else.

## Cron

Five numeric fields, validated in the SPI carrier so the subset is a property of the type rather than a rule drivers are trusted to follow. Name aliases (`MON`, `JAN`) are rejected: implementations disagree on whether day-of-week `0` is Sunday or Monday, and the alias spelling hides the disagreement.

Interpreted in **UTC** — local time would expose the two classic DST defects, a daily job running twice on the autumn overlap and one silently skipping the spring gap. Day-of-month and day-of-week are **OR'd** when both are restricted; that is standard cron and it surprises people, so it is stated in the contract and tested.

## Refactor under lint pressure

PMD flagged the scheduler at complexity 46 with too many methods. That was real signal, not noise — it was doing queue management, state transitions, trigger arithmetic, context rebinding and thread lifecycle at once. Split into `CommunityJobRegistry` (guarded state, so the locking is auditable in one place), `CommunityTriggerClock` (trigger → deadline, where the two clocks meet), and `CapturedContext` (the identity pair, which travels together). One suppression added, on the job-body catch, with the reason inline.

## Verification

Full reactor `mvn clean install` green with lint gates active, no skip flags.

| Suite | Cases |
|---|---|
| `AbstractJobSchedulerTck` | 18 (Triggers 6, Lifecycle 7, Identity 3, Failure 2) |
| `CronSyntaxTest` | 33 |
| `CommunityCronScheduleTest` | 12 |
| `CommunityJobJfrTest` | 3 |
| `CommunitySchedulingArchitectureTest` | 4 |
| `ExerisArchitectureTest` | 13 (was 12) |

Bootstrap wiring — the `KernelProviders` slot pair, config keys, lifecycle integration — is S9.


arkstack-dev:
Blocking finding closed, plus findings 2-3 and the Sonar batch. Pushed as `777cddda`.

## 1. Shutdown-drain race — real, and I verified the trace

`dispatchLoop` checked a job out of the queue in one critical section and registered its worker in another. Between them the job was in **neither** structure, so a `close()` landing there snapshotted past it and returned while the job was still about to run.

Check-out and worker registration are now **one locked step**, so a job is at every moment either queued or in flight. One thing your fix sketch didn't cover, which I hit while testing: `close()` can join a worker that has been registered but **not yet started**, and `Thread.join()` on an unstarted thread returns immediately. So the worker also re-checks the running flag before executing — that covers whichever side of `start()` the race falls on.

### I could not reproduce it, and I'm not claiming otherwise

You were right that `closeDrainsInFlight` can't reach the window. I added a case that races a time advance against `close()` — but **reintroducing the two-critical-section shape does not fail it**. I ran that mutation: 19/19 still green. The interleaving needed is a few instructions wide and isn't reachable by racing two threads from outside.

So the race is closed **by construction, not by detection**. The new TCK case asserts the coarser property (`close()` must not return while a body it started is unfinished) and its Javadoc states plainly that it does not reproduce the original window — I'd rather that be written down than have someone later read a green run as proof the window is guarded.

A process note while I was there: my first attempt at that mutation left the **broken version in the working tree**, and I didn't notice until PMD flagged the leftover on the next full build. The verification numbers below are from a run after restoring and re-checking the invariant is present in the file.

## 2. Dead factory and mismatched code docs

Both correct. `noCapturedContext()` is removed. `EX-JOB-9001` and `EX-JOB-9003` never reach a thrown `rawArgs`, so their registry entries now say they are JFR-only and name the event's fields — including `jobId`, which the old layout comment omitted — instead of describing a layout that doesn't exist. `EX-JOB-9002` keeps its factory and its layout.

## 3. POM indentation

Fixed.

## Sonar

| Finding | Resolution |
|---|---|
| `ScopedValue.Carrier` never consumed | Three explicit chains, each terminating in its own `run()`, instead of an accumulated nullable carrier |
| `await`/`awaitNanos` outside a loop; return value ignored | Both clocks now wait in condition loops against a signal counter — a spurious wake-up neither ends a wait early nor restarts the full interval |
| Catch `Exception`, not `Error` | Taken. `Error` propagates; `runJob`'s outcome starts at `FAILED`, so an escaping `Error` still retires the job as failed rather than as completed — which the previous shape got wrong |
| `minutes`/`MINUTES` clashes | Field-size constants renamed (`MINUTE_COUNT`, …) |
| Record patterns, `continue` count, unnamed catch patterns | Applied |
| `record()` matches a restricted identifier | Renamed `recordEvents` |
| `Thread.sleep()` in a test | Replaced with a bounded state poll on the handle — returns as soon as the refusal is recorded instead of paying a fixed delay every run |

**Left alone deliberately:** `java:S2187` ("Add some tests to this class") on the two `@Nested`-only classes. That fires on the house test shape across this repo, and restructuring a test to satisfy it trades a real organising principle for a linter's file-shape preference.

## On the O(n) observation

Agreed it's a §C signal, not a gate. Worth adding: the linear scan is over the *queued* set only — a job is out of the queue for the whole run — so it's bounded by scheduled jobs, not by concurrency. A delay-ordered priority queue is the natural upgrade if that ever grows.

## Verification

Full reactor `mvn clean install` green with lint gates active. `AbstractJobSchedulerTck` **19** (was 18), `CronSyntaxTest` 33, `CommunityCronScheduleTest` 12, `CommunityJobJfrTest` 3, `CommunitySchedulingArchitectureTest` 4, `ExerisArchitectureTest` 13.

PR #263, opened 2026-07-31
docs(roadmap): two dogfooding-surfaced SPI gaps — graph multi-hop traversal, flow await

arkstack-dev:
Records two SPI gaps surfaced by downstream use. Docs-only; no kernel commits.

Both **verified against the tree** before writing — neither had an existing ROADMAP entry.

## Graph: heterogeneous multi-hop traversal

`GraphTraversal` carries exactly one `GraphEdgeDescriptor`, and every `GraphSession` entry point that consumes a traversal has that shape (`traverseBreadthFirst`, `streamBfsJson`, `findShortestPath`). A two-hop `User -[PURCHASED]-> Product -[SIMILAR_TO]-> Product` is not expressible as one request.

The cost is more than ergonomic: the join moves into application code (one round trip per hop, N+1 on the second), the intermediate set lands on the heap — against the contract `streamBfsJson` otherwise honours through `LoanedBuffer` — and second-hop tenant scoping leaves the engine.

RFC-first, because one legitimate option is **declining the primitive**: arbitrary path expressions may be a query-language problem the kernel deliberately does not own. Whatever lands must define how depth interacts with hop count and keep `StorageContext` scoping engine-side on every hop.

**post-1.0** — graph is outside the six-subsystem 1.0 core, and this is contract widening rather than a correctness fix.

## Flow: no way to await a flow

`FlowScheduler.schedule` returns `void`; neither it nor `FlowEngine` exposes a completion surface. Request/response over a saga — answer the HTTP call when it settles — has no kernel answer, so every consumer invents its own correlation and waiting.

RFC-first because the hard part is not "add a future": it is what completion *means* for a durable flow that can park across a restart the awaiter does not survive. That is exactly where a naive API would mislead. Constrained to no `CompletableFuture` and no new `StructuredTaskScope` site.

**1.0-recommended** — flow *is* in the 1.0 core, and "replaces the orchestration layer" is one of the two load-bearing product claims; a flow nobody can wait on weakens it. Sequenced behind the v0.11 flow-versioning and continuity work.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -1961,7 +1961,9 @@ See also: v0.11 §"Transport: PAQS Execution-Seam Port (M1)" (seam landed on the
 
 ### Transport: Connection Cap Refuses Silently (surfaced by load-test triage, 2026-07-31)
 
-**Gap:** `NativeTcpCarrier`'s accept loop reserves a slot against `TransportConfig.maxConnections()` (default `HttpConfig.DEFAULT_MAX_CONNECTIONS` = 1000, `http.maxConnections`). When the cap is reached it calls `closeQuietly(currentChannel)` and continues: the connection is **accepted at TCP level and then closed with no log line, no JFR event, and no counter an operator can read**. From the client the socket opens and dies; from the server nothing happened.
+**Gap:** `NativeTcpCarrier`'s accept loop reserves a slot against `TransportConfig.maxConnections()` (default `HttpConfig.DEFAULT_MAX_CONNECTIONS` = 1000, `http.maxConnections`). When the cap is reached it calls `closeQuietly(currentChannel)` and continues: the connection is **accepted at TCP level and then closed with no log line and no JFR event**. From the client the socket opens and dies; from the server nothing happened.
+
+Sharper than "no counter", which is how this was first written: `TransportStats` already carries `totalRejected` — *"cumulative number of connections/streams rejected (load shedding)"* — and the carrier fed it from the PAQS load-shedder alone. The one field an operator consults when asking whether the server is turning work away therefore read **zero while every connection was being refused**. That is worse than absent telemetry, because a zero is read as evidence the fault lies elsewhere.
 
 This is the only refusal path in the kernel with no telemetry at all, which makes it the hardest failure to diagnose and the easiest to misattribute. A load test that trips it sees connection-level errors with a clean server log and will reasonably conclude the fault is elsewhere — a shape that has already cost triage time (see below). It also breaks the Glass-Box premise: every other shed or deny in the kernel emits an event, and admission decisions in persistence emit `AdmissionDecisionEvent` specifically so operators can attribute them.
 
@@ -1974,6 +1976,8 @@ Two properties make this cap easy to hit unexpectedly: it counts **concurrent co
 **Merge Gate:** Refusal event registered in `docs/subsystems/telemetry.md` §Required Events and asserted by a driver-local JFR test that drives the cap to its limit; transport stats expose a monotonic refusal counter; `docs/subsystems/transport.md` documents the cap, its default, and what a client observes when it trips.
 
 **1.0 disposition:** 1.0-recommended. Transport *is* in the 1.0 core, and an undiagnosable refusal path is the kind of thing that turns a support conversation into an accusation.
+
+**Status (v0.11):** observability half **DELIVERED** — `CommunityConnectionRefusedEvent` per refusal, and `totalRejected` now sums both refusal paths. The policy half is open and deliberately unbundled: whether an accept-time cap is the right mechanism against request-level shedding that can answer with a status, and whether 1000 is the right default for the reference deployment.
 
 ---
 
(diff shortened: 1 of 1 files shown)


arkstack-dev:
Added a third entry, found while triaging the load-test anomaly.

`NativeTcpCarrier`'s accept loop reserves a slot against `maxConnections` (default **1000**) and on failure does:

```java
slotReserved = tryReserveConnectionSlot();
if (!slotReserved) {
    closeQuietly(currentChannel);
    acceptedChannel = serverChannel.accept();
    continue;
}
```

Accepted at TCP level, then closed with **no log line, no JFR event, no counter**. It is the only refusal path in the kernel with no telemetry at all — persistence emits `AdmissionDecisionEvent` for precisely this reason.

The cap counts **concurrent connections, not rate**, so a transient overlap of two client phases can cross it at a request rate that is otherwise comfortable.

The resolution deliberately separates the two questions: the silence is the defect to fix; whether 1000 is the right default, and whether an accept-time cap beats request-level shedding where the response can carry a status, is a policy decision that should not ride along in the same commit.

arkstack-dev:
Both nits addressed, plus one error the review didn't catch.

**Placement — accepted.** All three entries carry a 1.0 disposition, which is the Road-to-1.0 schema. They sat at the tail of v0.12 by proximity, not intent, and none of them is planned for v0.12. Moved.

**"six-subsystem" — the source exists.** Road to 1.0 names them: `transport / http / security / persistence / flow / events` (+ `memory` foundation), and closes with "unbreakable on six things". I've dropped the figure anyway, because of the next item.

**The error underneath that nit.** Chasing the citation surfaced something worse than an unsourced number: two paragraphs below the six-subsystem line sits **"Graph in 1.0 — DECIDED: stays in 1.0"**. My entry justified its post-1.0 disposition with *"graph is not in the 1.0 core"* — which contradicts a decision this document explicitly records as closed.

The disposition itself is right; the reasoning was not. Graph the subsystem **is** in 1.0 and is TCK-backed (GRAPH-111). What is post-1.0 is *this contract widening* — new traversal surface the 1.0 ruling never assessed, and adding it inside the window would re-open a scope question that was deliberately shut. The entry now says that.

Worth noting for the next doc review: the nit that looked cosmetic was the one that led to a real contradiction, and the two claims that looked substantive both verified clean on the first pass.

arkstack-dev:
Blocker corrected, plus two findings added. Pushed as `13835ea0`.

## The DELIVERED claim — right about the effect, wrong about the cause

`CommunityConnectionRefusedEvent` **does** exist, and `totalRejected` **does** sum both refusal paths — in **#264**, which this branch does not carry. Your grep was against this branch, where it legitimately isn't present.

That distinction doesn't rescue the claim. **DELIVERED in this document means present on `development/*`**, and I wrote it while the delivery was still in review — which is the same unverifiable-claim class you caught last round, just pointing at a shipped fix instead of a figure. Reworded to say the observability half is implemented in a separate change and not yet on the development branch, which stays accurate under either merge order.

Worth noting the review method held here: checking the branch rather than the repo-at-large is what surfaced it. A cross-branch grep would have found the class and let a premature claim through.

## Two findings added, split by the disposition they carry

The user asked for a sweep of hardcoded limits and other silent drops. Both results are now in the document.

### v0.11 — accept-loop `RuntimeException` swallowed whole

```java
} catch (RuntimeException exception) {   // bound, never used
    closeQuietly(currentChannel);
}
```

Anything thrown while setting up an accepted connection is discarded: no log, no event, no counter.

Different in kind from the `maxConnections` refusal already recorded. That one hid a **policy** decision at a known limit; this hides **defects** — an allocator exhaustion or TLS init failure looks exactly like a healthy server under no load, and a *repeating* fault is indistinguishable from an intermittent network problem. The sibling paths in the same file close the same resources and **rethrow**; only the accept loop swallows.

The resolution deliberately does **not** change recovery: continuing the loop is correct, and conflating "make it visible" with "make it fatal" would trade a silent drop for an availability regression.

### 1.0-blocking — operational limits with no configuration path

**PAQS has no config surface at all.** `AdmissionController` takes no config object, and neither the transport scheduler nor the events package reads the config provider once — no `paqs.*` or `events.*` key exists. So `MAX_ACTIVE_STREAMS = 5000`, which its own Javadoc says sheds *"regardless of memory pressure"*, cannot be raised, lowered, or disabled by any means.

The sharpest row is an asymmetry: `http.maxRequestHeaderSize` and `maxRequestHeaderCount` are honoured on HTTP/1 and **silently stop applying once a client negotiates h2** — the HTTP/2 equivalents are compile-time constants. Nothing on the config surface says the knob is version-scoped.

A second class is worse than hardcoding: four settings read from `System.getProperty` — outside the config provider, not hot-reloadable, invisible to `KernelConfigRegistry`. They *look* configurable to whoever wrote them and are undiscoverable to whoever operates them.

The substantive part of the resolution is not the mechanical promotion — it is deciding **once** what "disable" means for a protective limit. An admission controller with no ceiling is a legitimate configuration for a JVM-controlled deployment and a foot-gun for a shared one, and the contract should say which it supports.

Protocol invariants — HPACK table shapes, status-code ranges, UTF-8 boundaries — stay hardcoded and are named out of scope.

PR #264, opened 2026-07-31
fix(transport): make an accept-time connection refusal observable

arkstack-dev:
Makes the accept-time connection refusal visible. Behaviour unchanged; only its observability is new.

## It was worse than "silent"

`NativeTcpCarrier` reserves a slot against `maxConnections` (default **1000**) and, at the ceiling, accepts the connection at TCP level and closes it — no log, no event.

But `TransportStats` **already carries `totalRejected`**, documented as *"cumulative number of connections/streams rejected (load shedding)"* — and the carrier fed it from the PAQS load-shedder alone:

```java
long rejected = 0L;
if (localPaqs != null) {
    rejected = localPaqs.loadShedder().shedCount();
}
```

So the one field an operator consults when asking *"is this server turning work away"* read **zero while the carrier refused every connection**. That is worse than absent telemetry: a zero is read as evidence the fault lies elsewhere, which is exactly how a transport-level refusal gets misattributed for days.

## What landed

- **`CommunityConnectionRefusedEvent`** per refusal, with the bind address, active count, configured ceiling, and a running total whose *slope* distinguishes a blip from a wall. No peer identity — the refusal happens before the remote address is resolved, so there is nothing to record and nothing to leak.
- **`totalRejected` now sums both refusal paths.**

Per-refusal rather than aggregated, matching `StreamShedEvent`'s per-shed shape. The path only executes at the ceiling, so the event rate is bounded by the client's retry rate, not by served traffic — and a burst of these is the signal, not noise.

## What this deliberately does not decide

Whether an accept-time cap is the right mechanism at all — as against admitting and shedding at request level, where the response can carry a status — and whether 1000 is the right default. Both are open in the ROADMAP. Fixing the silence first is what lets those be taken on evidence instead of argument.

## Verification

Full reactor `mvn clean install` green with lint gates active.

`CommunityConnectionRefusalTest` drives a carrier with `maxConnections=2` past its ceiling and asserts both surfaces. Two mutations, both killing it:

| Mutation | Result |
|---|---|
| Drop the `recordRefusal()` call — back to the silent close | fails |
| Count the refusal but leave it out of `totalRejected` | fails |

The second matters as much as the first: it is the exact shape of the defect that existed, where a counter was maintained somewhere but the field an operator reads stayed at zero.

Documented in `docs/subsystems/transport.md` §"Connection ceiling" — including what a client observes — and registered in `telemetry.md` §Required Events.

Discharges the ROADMAP entry added in #263; the `Status` annotation there follows once that merges.


arkstack-dev:
All three addressed — the third taken further than raised. Pushed as `d4eda644`.

## 1. Stale SPI contract doc — fixed

You're right that this is the one an Enterprise binding author reads first. `totalRejected` now states the contract as **the sum across every refusal path an engine has**, and says why it matters: a driver counting only one path reports zero during a total refusal, which is read as evidence the fault lies elsewhere. That's the defect this PR fixes, stated where a second binding will encounter it.

## 2. Dead accessor — removed

## 3. I don't think this one is pre-existing

You filed it as a pre-existing gap and non-blocking. The gap in *coverage* is pre-existing — the transport TCK had no rejection cases at all. But **the obligation is created by this PR**: I changed what an SPI field means, and CLAUDE.md rule A makes changed observable SPI behaviour require `Abstract*Tck` coverage before merge. Waving it through on "the TCK never covered rejection anyway" would be using the absence of coverage as the reason not to add any.

So `AbstractTransportEngineTck` now drives an engine past a small ceiling and asserts `totalRejected` becomes positive, via a new `createEngineWithConnectionCeiling(max, port)`.

**Abstract, not defaulted-to-skip** — deliberately. A binding that never implemented it would pass the contract *by not running it*, which is exactly the defect shape here: the Community driver counted its accept-time refusals nowhere and reported a healthy zero. A skipping default makes a contract vacuous for precisely the bindings that never looked at it.

The suite supplies the port so it can drive clients without needing a binding-specific way to read the bound address.

Responsibility splits the same way as the blob slice: **the TCK asserts the contract** (`totalRejected` reflects every refusal path), **the driver-local test asserts the driver decision** (which JFR event, with which fields). Your point about an Enterprise binding honouring the same contract is exactly what the TCK half now enforces.

## Verification

Full reactor `mvn clean install` green with lint gates active.

New mutation: drop the ceiling refusal back out of `totalRejected` → `totalRejectedCountsCeilingRefusal` fails on *"a refusal the engine does not count is a refusal an operator cannot see"*. The three earlier mutations still hold.

Worth noting the `tryReserveConnectionSlot` CAS observation in your review is what makes the event's `activeConnections` snapshot trustworthy — I'd relied on it without stating it.

PR #266, opened 2026-07-31
fix(transport): re-land the accept-setup fault telemetry (#265 never reached development)

arkstack-dev:
**#265 shows MERGED but its content is not on `development/0.11.0`.** Re-landing it.

## What happened

I based #265 on #264's branch rather than on `development`. #264 was merged into `development` first, and #265 then merged into `fix/v011-transport-refusal-telemetry` — producing merge commit `a9851a77` on a branch that no longer feeds anything.

Ground truth at the time I checked: `CommunityAcceptFaultEvent.java` absent from the tree, and no accept-fault commit in `git log origin/development/0.11.0`. The PR list said MERGED; the tree said otherwise, and the tree was right.

This is my error in branch topology, not a review or merge mistake. Stacking a PR on an unmerged branch means the stacked change only lands if the base merges *last* — the opposite of the order these went in.

## Content

Unchanged from #265 — a clean cherry-pick of `f9b12137` onto current `development`, no conflicts:

- `CommunityAcceptFaultEvent` on the accept loop's `catch (RuntimeException)`, which previously bound the exception and never used it.
- Deliberately **not** in `TransportStats.totalRejected` — that field means work the engine *declined*, and a setup that broke declined nothing.
- Exception class only, never the message.
- Recovery unchanged; `transport.md` tables both accept-path failure modes side by side.

## Verification

`CommunityAcceptFaultTest` and `CommunityConnectionRefusalTest` green against the re-based tree. Full reactor build ran green on the original branch; the cherry-pick applied without conflict.


arkstack-dev:
Fixed in `25332449`. Each javadoc block now sits on the method it describes.

Worth naming the mechanism, because it's the second time this class of thing has bitten this branch chain: **a cherry-pick that applies with no conflict can still land a hunk somewhere that changes what a doc comment attaches to.** Git matched on surrounding lines; javadoc binding is positional. "Applied without conflict" is not the same as "applied correctly" — and I reported the former as if it were the latter in the PR body.

On the build claim: fair to flag rather than assume. Re-verified after the fix on this branch — `CommunityAcceptFaultTest` and `CommunityConnectionRefusalTest` green, and `mvn -pl exeris-kernel-community pmd:check checkstyle:check` clean. The full-reactor green in the original PR body was from the pre-cherry-pick branch, which is exactly the kind of carried-over claim your caveat is right to distrust.

PR #267, opened 2026-07-31
feat(scheduling): bootstrap wiring for the JobScheduler SPI (v0.11 S9, ADR-057)

arkstack-dev:
Closes Stream C. Bootstrap wiring for the scheduling SPI landed in #262.

## What landed

`SchedulingBootstrap` ranks `JobSchedulerProvider` through the shared `BootstrapProviderSelector`; `CommunitySchedulingSubsystem` binds both `KernelProviders` slots. Config key `scheduling.schedulerName`, read **through the config provider** — deliberately not a system property, since the limits sweep recorded four of those as a defect class and a new subsystem must not add a fifth.

`stop()` calls `JobScheduler.close()`, so the ADR-057 §6 drain is what shutdown **actually performs**, not a property the scheduler merely offers. The test asserts that by holding a job mid-run across `stop()`.

No provider on the classpath is terminal (`EX-JOB-9004`), not a degraded boot: a kernel that started with scheduling silently absent would let an application submit jobs that never run.

Selection commits `SchedulingBootstrapSelected`. Which provider won a ServiceLoader race is invisible afterwards and is the first thing anyone asks when a scheduler misbehaves — recorded rather than reconstructed from behaviour.

## The subsystem declares no dependencies, and that took a correction

My first attempt declared `List.of("memory", "security")`, reasoning that a job carries the identity that submitted it. **Wrong twice:**

1. There is no `security` subsystem in the registry — identity arrives through `ScopedValue` slots, not a boot-ordered component. `SubsystemInitOrderIntegrationTest` failed with *"depends on missing subsystem 'security'"*.
2. Even conceptually it was wrong: identity crosses this seam at **submit** time from the caller's own scope, which no boot order can influence.

The in-process driver allocates no off-heap memory, opens no connection, reads no store. Its construction is a config lookup and a virtual thread.

Worth noting where the error sat: one paragraph **above** the declaration, my own javadoc warned that "declaring a dependency it does not have would order the boot graph on a fiction". The existing boot-order suite is what caught it, not the reasoning next to it.

## Verification

Full reactor `mvn clean install` green with lint gates active. `CommunitySchedulingSubsystemTest` 4 cases; `AbstractJobSchedulerTck` 19 still green; boot-order suite green.

| Mutation | Result |
|---|---|
| Remove the drain from `stop()` | `stopDrains` fails |
| Suppress the selection event | `emitsSelectionEvent` fails |

Docs: `scheduling.md` gains a Bootstrap section (config key, dependency reasoning, terminal-no-provider, drain) and drops the "lands with the next slice" note; `telemetry.md` gains the selection event.


arkstack-dev:
Rebased onto `d55de297` and resolved. Conflict was in `docs/subsystems/telemetry.md` §Required Events — #266 and this branch both inserted a row before the same anchor line, so it was purely additive.

Both rows stay. I grouped them by subsystem rather than stitching them in merge order: `SchedulingBootstrapSelected` now sits with the three `JobDispatch`/`JobCompletion`/`JobFailure` rows, and `CommunityAcceptFault` sits next to `CommunityConnectionRefused` — the two transport rows are the contrast pair, and separating them would undo the point of the table entry.

Re-verified after the rebase rather than assuming: full reactor `mvn clean install` green with lint gates active, `CommunitySchedulingSubsystemTest` 4, and both transport telemetry tests from #264/#266 still green on the merged result.

PR #268, opened 2026-08-01
feat(storage): S3-compatible blob driver closes the two-binding gate (v0.11 S6, ADR-056)

arkstack-dev:
Second in-repo binding of the blob contract, against a MinIO-compatible endpoint. This is the slice the ≥2-binding gate was waiting for: a contract with one binding describes that binding, and a filesystem and an object store disagree about nearly everything below `BlobStore`, so a rule both satisfy is a rule about blobs.

## Scope, stated rather than left to folklore

Recorded as an **ADR-056 §10 amendment** — the original obligation said "the supported authentication subset is documented with the driver", and these are the four things building it settled:

| | |
|---|---|
| **Signing subset** | Header signing over `host` / `x-amz-content-sha256` / `x-amz-date`, plus query-signed presigned URLs. No chunked (`STREAMING-*`) payload signing, no session tokens, no signing caller-supplied `x-amz-*` headers — each serves a capability this driver does not have, so implementing it would be signing for requests it cannot make. |
| **Cleartext only, refused not downgraded** | `CommunityHttpTransportFactory` wires certificate material for listeners only, so a `CLIENT`-mode engine speaks cleartext whatever the scheme says. An `https://` endpoint is **rejected at construction**. Sending SigV4 credentials in the clear because a scheme was ignored is not a failure that may be quiet. |
| **The ceiling is a named knob** | `s3.maxObjectBytes`, default 8 MiB. Without multipart the object is held in one buffer, so a ceiling exists whether or not it is named; naming it makes the refusal loud (`EX-BLOB-8005`, before any allocation or request) instead of surfacing as an allocation failure. |
| **Two round trips per read** | Every download begins with a `HEAD`. The driver must know an object's size before deciding to pull it into one buffer, and a ranged read needs the total for `BlobMetadata.sizeBytes()` anyway. `DELETE` pays it too: S3 answers `204` regardless, and the contract promises the caller can tell a removal from a no-op. |

## Prerequisite fix: the client could not issue HEAD

`CommunityHttpClientResponseDecoder` read `Content-Length` as a promise of bytes to come, so a `HEAD` response — which declares the object's length and sends no body — ended as `Truncated HTTP response body: expected N bytes but received 0`. The method was unusable, found the moment the driver tried to `stat` without downloading.

Framing now follows the request method (RFC 9110 §6.4.1), threaded from the engine because a response cannot carry it. `CommunityHttpClientHeadResponseTest` pins it in the default build, including the negative direction — the same bytes read as a bodied response still fail, so the flag is demonstrably what makes `HEAD` work.

## Why the filesystem driver is touched

`CommunityBlobFailures` was a static holder with `"ExerisCommunity/FilesystemBlob"` baked in. A second driver made the provider name a per-driver fact, and the two honest options were passing it at every one of ~20 call sites or binding it once. Bound once, a driver cannot attribute a failure to its sibling through an argument slip, and both drivers keep the same event names — filter `eu.exeris.kernel.storage.*` and read `providerName`.

New: `BlobCeilingExceeded` (a refusal by policy is not a failure, so it is not a `BlobTransferFailed` with a special reason) and a `remoteStatus` field on `BlobTransferFailed` (`403` is the caller's credentials or clock, `5xx` is the store's problem — the one field that decides who has to act).

## Verification

- `mvn clean install` green on the **full reactor**, lint gates active, no skip flags.
- `ExerisArchitectureTest` 13/13.
- `CommunityS3BlobStorageTckIT` 18/18 against live MinIO — **re-run after the refactors**, not only before them.
- **Non-vacuity by mutation:** a wrong secret key fails 15 of 18. The three survivors are exactly the cases that need no verified signature — local TTL validation, the isolation deny, and the upload length check. The suite genuinely depends on SigV4 verifying against a real store.
- Default build gains 19 cases (`CommunityS3SignerTest`, `CommunityS3ObjectKeyTest`, `CommunityHttpClientHeadResponseTest`). The integration gate proves a round trip works; it cannot prove *why* two tenants stay apart, because a driver concatenating the isolation key verbatim would round-trip just as happily.

**CI:** the IT rides the existing community integration job, which already runs every `@Tag("integration")` test in this module (the Keycloak suite has been there since v0.10). A second job selecting the same tag in the same module would run the same tests twice — ADR-056's "with its own CI gate" wording is amended to say so.

## Recorded, not closed

Two ROADMAP entries rather than silent gaps:

1. **Two providers, no way to choose.** Both are registered at the same Community priority and nothing loads `BlobStorageProvider` through `ServiceLoader` yet — inert today, but the moment a storage subsystem bootstraps, discovery order would decide where a tenant's objects land. Resolution sketched against the scheduling precedent, with one addition: with two providers at equal priority the choice must be *stated* by config, and an ambiguous configuration must fail at startup rather than pick.
2. **Every response buffer is sized for the largest possible one.** `resolveAggregateCapacity()` derives once from `maxRequestBodyBytes`, so an 8 MiB-ceiling engine allocates 8 MiB to read a 200-byte `HEAD`. Not a correctness bug — an oversized response is refused loudly — but it is per-request waste, and the knob is a *server request* limit by name being reused as a *client response* ceiling. Separate client-side knob tagged 1.0 (configuration surface); per-response sizing follows.


arkstack-dev:
Fixed in `ed294eee`. The finding was right, and right about the reason: the knob's entire purpose is to turn an implicit ceiling into a loud refusal, and leaving it unbounded reintroduced the same failure one level up — a transfer passing its own limit check and then allocating from a wrapped `int`.

**Bounded at construction**, not at the transfer. A ceiling that cannot be honoured is a configuration error, and the point of naming the knob was to fail at a knowable moment rather than on the first large upload in production.

`MAX_SUPPORTED_OBJECT_BYTES` subtracts **both** headrooms that sit on top of the object, not just this driver's: the 64 KiB it adds itself, and the 8 KiB framing allowance the client engine adds to whatever ceiling it is handed (`resolveAggregateCapacity`). Subtracting only the first would still overflow at the maximum, since the engine's sum is what actually gets narrowed. That second constant is named and commented in `CommunityS3Settings` precisely because it is a coupling to another class's internals — if the engine's allowance ever changes, that is where to look.

## `CommunityS3SettingsTest` (new)

The finding exposed a coverage gap wider than itself: the parse path had **no default-build test at all**. Only the MinIO gate touched it, and only on the happy path — so the `https` refusal and the required-property messages were as untested as the ceiling. Eight cases now cover the bound, a concrete 4 GiB attempt (the shape an operator reaches for after reading "raise this to the largest object a deployment genuinely stores"), the largest accepted ceiling still leaving the engine's buffer addressable, the `https` refusal, and which key a missing credential names.

**Non-vacuity by mutation:** disabling the bound fails exactly the two ceiling cases and nothing else — the other six stay green, so they are not passing on the guard's coattails.

## Docs

`storage.md` and the ADR-056 §10 amendment both state the upper bound and why it exists, since the doc's "raise this" advice is what makes the hole reachable.

## Verification

- `mvn clean install` green on the full reactor, lint gates active, no skip flags
- `ExerisArchitectureTest` 13/13
- `CommunityS3BlobStorageTckIT` 18/18 against live MinIO — re-run after the change, not carried over

On the two accepted observations: the `delete()` TOCTOU reading matches mine (the contract promises distinguishing a removal from a no-op, not atomicity across concurrent callers — and the filesystem driver has the same window), and the `remoteRefused`/`exceedsCeiling` naming was deliberately kept vocabulary-free so a second object-store driver does not inherit S3's dialect through the error registry.

PR #269, opened 2026-08-01
feat(http): stream routes accept path templates, so a generated per-action route can match (v0.11 S20)

arkstack-dev:
Closes the Stream I entry. The streaming table was an exact-match `Map` while the tooling generator emits **templated** stream paths (ADR-044 Slice 2 — `<base>/{id}/actions/<kebab>`), so every per-action stream route registered successfully and then never matched a concrete request. A dead route that 404s on a real boot, accepted by the builder without a word. Not yet user-visible only because the per-action driver is still a keep-alive scaffold — which is why this is a v0.11 slice and was never a v0.10.x hotfix.

## Both halves, together

| | |
|---|---|
| **Templates** | The streaming table compiles `{name}` paths through the **same** `PathTemplate` the respond-once table uses, with the same exact-before-template precedence. One compiled type serves both tables, so they cannot drift into disagreeing about what `/x/{id}` means — which is exactly how this defect would come back. |
| **Fail-fast** | A malformed brace throws at `Builder.streamRoute`. A silently-dead stream registration is now unrepresentable — the property the ROADMAP made mandatory under *either* shape decision. |

Taken as the default direction without waiting on the tooling EV1-stream ruling: it subsumes the fail-fast property and stays correct under either outcome, as the alternative anticipated.

## `HttpStreamExchange.pathParams()`

A templated route that resolved and then dropped what it captured would leave the handler knowing a stream is open but not *what it is a stream of*. Delivered by `PathParamStreamExchange` — the streaming mirror of the existing `PathParamHttpExchange`: a decorator that adds one answer and forwards the rest, so routing state stays out of the wire engine. Only wrapped when a template captured something, so exact routes pay nothing.

## Incidental allocation win

`PathTemplate` splits `matches()` from `capture()`. A miss — the common case when a router walks its template list — now allocates **no map at all**, where the previous shape built one and threw it away. A hit walks the segments twice: a handful of comparisons, against an allocation per miss.

## Verification

- `mvn clean install` green on the **full reactor**, lint gates active, no skip flags
- `ExerisArchitectureTest` 13/13
- `HttpRouterTest` 38 cases (`StreamingRegistration` 3 → 9)
- `AbstractHttpStreamExchangeTck` gains a `PathParameters` case, green on the Community binding: `pathParams()` is never null and is **immutable** — the map is routing state, so a handler able to write to it could change what a later request resolves to
- **Non-vacuity by mutation:** registering every stream path as exact fails 4 of the 6 new router cases. The two survivors are the negative ones (wrong shape, table disjointness) — they assert that *nothing* matches, which is precisely what a registration bug cannot break

## Call-site change

`HttpRouter.resolveStream` now returns a `StreamMatch` (handler + captured params) rather than a bare handler, since a template hit has two things to report. `StreamMatch.exact(handler)` covers the no-params case; the only production caller is `CommunityHttpStreamDispatcher`.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -1750,6 +1750,16 @@ See also: ADR-034 (`KernelWebClient` facade, superseding ADR-026), ADR-032 (`Htt
 
 **Owner:** HTTP subsystem (router); shape decision shared with `exeris-tooling` (ADR-044 EV1-stream slice).
 
+**Status (v0.11): DELIVERED — kernel-side template matching.** The streaming table now compiles `{name}`
+paths through the same `PathTemplate` the respond-once table uses, with the same exact-before-template
+precedence, and captured values reach the handler through the new `HttpStreamExchange.pathParams()` (a
+`PathParamStreamExchange` decorator, mirroring `PathParamHttpExchange`). One compiled template type
+serves both tables, so they cannot drift into disagreeing about what `/x/{id}` means. The fail-fast half
+landed with it: a malformed brace throws at `Builder.streamRoute`, so a silently-dead stream registration
+is unrepresentable. This is the default direction below, chosen without waiting on the tooling ruling —
+it subsumes the fail-fast property and stays correct under either tooling outcome, exactly as the
+alternative anticipated.
+
 **Resolution:** Rule the shape inside the tooling EV1-stream slice, then land the kernel side in v0.11. Default direction — kernel-side template matching: extend the streaming table with the same template semantics the unary table already has (reuse `PathTemplateRoute` compile/match for `streamRoute` paths containing `{name}`; exact stream routes take precedence over template stream routes, mirroring unary precedence; captured params exposed on the streaming path). Alternative — ADR-044 amendment (tooling repo) redesigns per-action stream paths onto exact routes, and the kernel change collapses to a Javadoc clarification plus a build-time guard. **Either way, fail-fast becomes mandatory:** a `{` in a stream path must either resolve via templates or be rejected in `Builder.streamRoute` — a silently-dead registered route must become impossible.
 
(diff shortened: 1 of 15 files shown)


arkstack-dev:
Both nits addressed in `8c9c5379` — including the coverage gap, which I did not want to leave as a follow-up.

## The gap was worth closing here, not later

The finding is exactly right about its shape: every case proving templates work sat at the router level, and every Community call site built its match with `StreamMatch.exact(...)`, so the branch this PR actually adds — wrapping the engine in `PathParamStreamExchange` only when a template captured something — had nothing touching it. Neither did the decorator's delegation.

That is the same failure mode as the defect being fixed: a resolution that looks correct and never reaches production behaviour. Accepting "the router cases cover it" is the reasoning that let the original dead route ship in the first place, so it seemed like the wrong place to take the merge-gate's permission.

`CommunityStreamRouteTemplateTest` goes through `resolveStreamHandler` and `dispatchStream` with a real `HttpRouter`, and asserts the handler read `id=42` from its own `pathParams()` — plus that `request()` is forwarded rather than synthesised, since a decorator that quietly rebuilt the request would pass a params-only assertion. A second case pins the undecorated path: an exact route still reports an empty map, not a null one.

**Non-vacuity by mutation:** forcing the dispatcher to always hand over the bare engine fails the templated case and leaves the exact one green. That split is the right one — the exact path never wanted the decorator, so a mutation of the decorator branch must not disturb it.

## Nit 1 — shared predicate

Switched. It tested the same condition inline, but a second copy of "what counts as a template" is precisely the drift `PathTemplate` was extracted to prevent, so leaving it would have undercut the argument the PR makes for the extraction.

## Nit 2 — `PathParamStreamExchange` visibility

Your reading matches the intent: `PathParamHttpExchange` can stay package-private because `HttpRouter.handle()` wraps in the same package, while streaming dispatch happens in Community. Narrowing it would mean either moving streaming dispatch into Core or exposing the wrap some other way, both larger changes than the surface is worth. Flagged in the class Javadoc as a deliberate widening rather than left to be inferred.

## Verification

`mvn clean install` green on the full reactor with lint gates active; `ExerisArchitectureTest` 13/13; the new suite green and mutation-checked as above.

PR #238, opened 2026-07-09
chore(claude): expand CLAUDE.md into a full session operating guide

arkstack-dev:
## What

Five commits, one theme: make the `.claude` meta layer factually correct and self-sufficient so any model (Fable, Opus, Sonnet) runs this repo with the same discipline.

**1. `169ad8c` — CLAUDE.md expansion.** From rule list to full session operating guide: project overview, reactor module table with Wall import rules, hard constraints (TCK-before-merge, `KernelErrorCodes` single source), sharpened scoped bans, `LoanedBuffer` lifecycle, testing model (tagged tests excluded from default build, Windows TLS auto-skip), build/DoD section, branch/PR/release workflow, skill mandates.

**2. `d0dd01a` — review-round-1 corrections (verified against POMs).** The "lint footgun" was backwards: `pmd:check` binds to `verify` and `checkstyle:check` to `validate` since the first commit — the real footgun is skip flags poisoning the proof. ThreadLocal/Executors bans re-attributed to ArchUnit, not PMD. Same fixes applied to the `exeris-pr-preflight` skill.

**3. `282c972` — full skills audit + systems pass** (19 skills, 6 agents, 12 commands verified against repo ground truth):
- **Fixed:** `exeris-java26-panama-loom` (unconditional "prefer `StructuredTaskScope`" would have REJECTed the ROADMAP-mandated preview-clean migration), `exeris-subsystem-specialist` (10 → 13 modes), `exeris-performance-contract` + `exeris-architect-guardrails` (scope contradictions), `community-architecture-review` (stale pre-0.10 reality block).
- **Created** `exeris-release-integration` and `exeris-tagged-gate-runner` — the two highest-frequency repeated workflows with no encoding.
- **Added** CLAUDE.md "Operating Standards (Any Model)" — 9 mechanical rules from this repo's observed failure modes.

**4. `2a8e5ec` — review-round-3 corrections.** Continuity gate row fixed to `community` + `community-kafka` (matching the CI job's `-pl` list and kafka's default exclusions; core has zero continuity tests), release-notes pattern fixed to `vX.Y.0-release-notes.md`.

**5. `1796c2c` — triage merge (maintainer-approved).** Six micro-skills (task-classifier / risk-prioritizer / routing-planner / validation-gate-planner / multi-agent-handoff / subsystem-scope-detector) consolidated into one `exeris-triage` skill with a strict-order output contract; references updated in `exeris-router`, `exeris-pr-review-waste-hunter`, and CLAUDE.md.

## Why

[...]

PR #271, opened 2026-08-02
test(transport): drive the Berkeley socket seam over real loopback (v0.11 S14)

arkstack-dev:
Stream F housekeeping. Writes the round-trip gate the plan listed as a loss risk — the file was untracked and is gone from the tree and from git history, so this is from scratch, not a recovery — and closes the two v0.6 transport entries that had no disposition.

## The gap

`CoreSyscallLoaderTest` and `SyscallHandlesTest` prove every symbol resolves and no handle is null on its platform. Nothing invoked one. The build could not distinguish a working socket seam from handles that merely exist, and the C-type→`ValueLayout` table `CoreSyscallLoader` documents had no executable check at all.

`SyscallLoopbackRoundTripIT`: socket → bind → listen → connect → accept → send → recv over 127.0.0.1, byte-exact comparison on the far side, plus a refusal case.

## Mutation results — including the one that survived

| Mutation | Result |
|---|---|
| `recv` bound to the `send` symbol | **Caught** — and only by the content assertion. Both counts still read 15; the buffer came back all zeros. The byte comparison is load-bearing, not decorative. |
| `send`'s `size_t` as `JAVA_INT` instead of `JAVA_LONG` | **Survived.** x86-64 SysV zero-extends the argument register, so a small length is indistinguishable either way. Integer-width entries in that table stay *reviewed-not-tested* on this platform — the Javadoc states this rather than implying coverage the test does not have. |
| `connect` bound to the `bind` symbol | **Exposed the refusal case as vacuous.** "Correctly refused" and "broken beyond use" both return `-1`, so it passed while the round-trip failed. |

That third result is why the refusal case looks the way it does now: it asserts **both** outcomes through one handle — a successful connect to a live listener immediately before the refused one. Re-running the same mutation now fails both cases, the refusal on its positive control. Without the positive control the assertion could not tell a working `connect()` from one that fails on everything.

All three reverted; the loader is byte-identical to HEAD.

## Two deliberate departures from the plan text

**No `@Tag("integration")`.** Core's pom binds Failsafe over `**/*IT.java` at `integration-test`/`verify`, so the name alone puts the test in the default build — confirmed by watching it execute inside `mvn clean install`, not inferred from the pom. Tagging it would drop it from Surefire opt-out paths (`-DexcludedGroups=integration`) while Failsafe ran it anyway: the exact mismatch removed from `OffHeapTlsEngineLoopbackIT` in v0.8 Sprint 6.

**macOS excluded**, on a concrete layout difference rather than caution: BSD `sockaddr_in` opens with a one-byte `sin_len` before `sin_family`, so the shared 16-byte address would be read with the family in the wrong place. Writing a second, unexecuted layout would be guessing.

## ROADMAP dispositions

**MpscArrayQueue ingress handoff → DISCHARGED.** Done by PERF-063 under a different name, which is why nobody closed the row: JCTools `MpscUnboundedArrayQueue` in `NativeTcpReactor`, not Agrona in `NativeTcpCarrier`. Both merge-gate halves exist (pinning TCK + its 2/4-reactor variants, and `NativeTcpTransportStressTest` behind `transport-stress-gate`). Recorded with the deviations so the row is not re-opened by someone diffing wording against code. `transport.md`'s stale Agrona attribution corrected to match.

**FFM socket migration → PARTIAL.** The validation seam is what landed. Migrating `NativeTcpCarrier` off NIO is a live-traffic change to the component every request crosses, and it should not begin while the Winsock half of the seam is unexecuted. **Precondition named and verified, not assumed:** every job in `.github/workflows/maven.yml` is `ubuntu-latest`, so no Windows runner exists and `loadWindows` has never run — here or in CI. That entry's own merge gate demands Winsock coverage before milestone close, which no v0.11 slice could satisfy.

## One repo-hygiene fix

This branch's own `exeris-pr-preflight` skill still claimed *"`mvn verify` does not run `pmd:check` (it is `default-cli`). A green verify is NOT lint-clean."* The parent POM binds `pmd:check` to `verify` (`verify-complexity`) and `checkstyle:check` to `validate` (`validate-architecture`). PR #238 corrected this on `main` and it was never forward-ported, so the development line has been telling every session a false thing about its own lint gate — including the gate run for this PR. Corrected text ported.

**Related finding, deliberately not folded in:** #238 also expanded `CLAUDE.md` from 92 to 185 lines (build/verify definition-of-done, operating standards, scoped bans). That is on `main` only too. Porting a guardrail document of that size inside a test slice would be scope inflation, so it is flagged here rather than done silently.

## Verification

- `mvn clean install` — BUILD SUCCESS, `MAVEN_EXIT=0`, full reactor, no skip flags; the new IT ran inside it (2/2).
- `ExerisArchitectureTest` — 13/13.
- Mutations reverted, loader confirmed clean against HEAD.
- Relative links in changed docs resolve.

Test-only plus docs: no SPI surface, no observable behaviour change, no ADR.

### DIFF
--- a/.claude/skills/exeris-pr-preflight/SKILL.md
+++ b/.claude/skills/exeris-pr-preflight/SKILL.md
@@ -1,12 +1,12 @@
 ---
 name: exeris-pr-preflight
-description: Pre-PR gate for Exeris Kernel — run BEFORE opening or pushing any PR. Enforces the lint footgun (mvn verify SKIPS PMD), the architecture guard test, fresh-branch-per-task discipline, and the register-ADR-before-claiming rule. Use whenever about to commit/push a branch or open a PR in the kernel repo.
+description: Pre-PR gate for Exeris Kernel — run BEFORE opening or pushing any PR. Enforces lint-gate integrity (skip flags poison a green build's lint proof), the architecture guard test, fresh-branch-per-task discipline, and the register-ADR-before-claiming rule. Use whenever about to commit/push a branch or open a PR in the kernel repo.
 ---
 
 # Exeris PR Preflight
 
 ## Purpose
-Catch the recurring, expensive footguns that a green `mvn verify` does **not** catch, before a PR is opened. This is a workflow skill (it runs checks), not a review-lens.
+Catch the recurring, expensive footguns that a green build alone does **not** prove are absent (lint skipped via flags, ignored arch rules, tagged-test gaps), before a PR is opened. This is a workflow skill (it runs checks), not a review-lens.
 
 ## When to Use
 - About to push a branch or open a PR in `exeris-kernel`.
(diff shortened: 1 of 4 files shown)


arkstack-dev:
Nit fixed in `dd0e733a`.

**Verified at the source before changing anything.** `SyscallHandles.java:60-61` does require `WSACleanup` before the owning arena closes, and `CoreSyscallLoader.loadWindows` calls `WSAStartup` on every `load()`. These tests load into a fresh `Arena.ofConfined()` per method, so on Windows each method would have closed its arena on an unbalanced refcount. `releaseWinsock(handles)` now runs last in both `finally` blocks, after the descriptors close.

**One correction to the framing, in the fix's favour.** `CoreSyscallLoaderTest` does not have this problem — not "has it too, pre-existing". It loads once into `Arena.global()`, which never closes, so the documented clause never engages and its single startup is process-lifetime. This IT was the first to genuinely violate the contract, which is why the fix belongs here and correctly does not spread to the sibling.

**Stated limit.** This branch is never taken in this repository: no Windows runner exists, as the FFM-migration disposition in the same PR records. The change satisfies the documented contract rather than a green run, and the Javadoc says exactly that instead of implying coverage. The Linux build confirms no regression, not the fix itself.

Verification after the change: `mvn clean install` BUILD SUCCESS, `MAVEN_EXIT=0`, full reactor, no skip flags, IT 2/2 inside it; `ExerisArchitectureTest` 13/13.

On the independent re-verification of the POM bindings — worth noting the correction runs in the direction that matters. The old skill text claimed a green `verify` was *not* lint-clean, which would push a session into redundant checks and into distrusting sound evidence. Confirming `verify-complexity` really is bound to `verify` closes the myth in both directions.

PR #272, opened 2026-08-02
fix(transport): drain in-flight streams before teardown, not after

arkstack-dev:
Surfaced from a downstream flaky test whose failing assertion turned out to be the correct one. `TransportEngine.stop()` has promised graceful drain since 0.5.0; the Community carrier ran every step of it, in an order that made the drain inert.

## The defect

`TransportEngine.stop()` (SPI, not just prose):

> Initiates a graceful stop — no new connections accepted, **in-flight streams are drained for a bounded period** before being forcefully closed.

`NativeTcpCarrier.stop()` did this:

```java
running.compareAndSet(true, false);   // reactors poll this → they exit
closeQuietly(serverChannel);          // ingress closed ✓
acceptor.join(2_000L);
for (r : reactors) r.wakeup();
for (r : reactors) r.join(2_000L);
closeSelectorAndChannels();           // every live channel closed
paqs.close();                         // ← the drain, against dead descriptors
```

`PaqsScheduler.close()` **is** the drain — it waits for the admission controller's active-stream count to reach zero under a 60-second hard deadline, exactly as `bootstrap.md` and the `transport.md` draining table describe. It just ran last. By the time it waited, the sockets those streams were answering on were gone, and so were the reactors: they poll `host.isRunning()`, and `stop()` clears that in its first line, so a single flag meant both *stop accepting* and *stop processing*.

Net effect: a request that reached its handler completed correctly and its response went nowhere. The peer saw a connection closed with no reply; an idempotent client retried and got a refusal, because ingress had closed several steps earlier.

## The fix

Ordering, plus the state distinction that was missing:

- New `draining` flag. Reactors poll `isReactorActive() == running || draining`, so they keep flushing responses for already-admitted streams while accepting none.
- `stop()` runs three phases: **close ingress** → **drain while the write path is alive** → **tear down**.

The 60-second deadline, the spin-then-yield backoff, the admission counting — all pre-existing and unchanged. Only *when* the drain runs is different.

## TCK — why this survived

The contract dates to 0.5.0. `AbstractTransportEngineTck` asserted only the state machine: `start()` → `stop()` → `close()`, idempotent restart, mode reporting. Nothing held a stream open across `stop()`, so a violation this visible passed every gate.

New `GracefulDrain` case does exactly that, plus a `createEngineWithHandler(port, handler)` factory so the suite can own the handler (the contract can't be asserted from outside it). Abstract rather than defaulted-to-skip, for the same reason the connection-ceiling factory is — a binding that silently didn't implement it would pass by not running.

**Non-vacuity by construction, no mutation required.** Against the unfixed carrier the case fails with `but was: []` — the client receives zero bytes. After the fix, 10/10.

**Cross-repo obligation:** any other binding of this TCK must implement the new factory.

## Telemetry

`CommunityTransportDrainEvent` — streams at start, streams remaining, duration. A non-zero remaining count means the deadline fired before the drain finished: the signal that in-flight work was severed, and the number to size `terminationGracePeriodSeconds` against. Until now a completed drain left no trace at all and a timed-out one left only a log line.

## Docs

`bootstrap.md` and the `transport.md` draining policy table were already correct — the code now matches them, so neither changed. Added a phase-order subsection and the JFR event to `transport.md`, including what the old order actually did.

## Verification

| Gate | Status |
|---|---|
| `mvn clean install`, full reactor, lint gates, no skip flags | `MAVEN_EXIT=0`, BUILD SUCCESS |
| `ExerisArchitectureTest` | 13/13 |
| Transport stress gate (`-DincludedGroups=stress -DexcludedGroups=`) | RUN+GREEN — 4 tests incl. multi-reactor pinning, which the reactor-predicate change directly affects |
| Community integration gate (`-DincludedGroups=integration -DexcludedGroups=`) | RUN+GREEN — 9 run, 3 skipped; all skips in the OpenSSL-conditional TLS loopback IT, untouched here. `NativeTcpCarrierIngressIntegrationTest` 5/5 |
| New TCK case, pre-fix | FAILS (`but was: []`) — the intended result |

### DIFF
--- a/docs/subsystems/transport.md
+++ b/docs/subsystems/transport.md
@@ -296,6 +296,28 @@ When PAQS sheds a stream or the Kernel initiates graceful shutdown:
 > **Why `FIN` not `RST` for load shedding?** `RST` causes immediate connection teardown on the client
 > side, which may interrupt in-flight retries and force the client to reconnect. `FIN` allows the client
 > to receive the `HTTP 503` response body, which is machine-readable and enables intelligent backoff.
+
+### Graceful-shutdown phase order (since 0.11)
+
+`NativeTcpCarrier.stop()` runs three phases, and the order is load-bearing rather than incidental:
+
+1. **Close ingress** — the listening channel closes and the acceptor joins. No new connections; the
+   reactors keep serving the ones already admitted.
+2. **Drain** — `PaqsScheduler.close()` waits for the admission controller's active-stream count to
+   reach zero under its 60-second hard deadline. The reactors are deliberately still looping here, so
+   a handler that completes during the drain can still have its response flushed.
+3. **Teardown** — reactors are woken and joined, then the selector and all remaining channels close.
+
+Until 0.11 the drain ran *after* teardown, which made it inert: handlers were waited on while their
+sockets were already closed, so an in-flight request completed and its response went nowhere. The peer
+observed a connection closed with no reply, and an idempotent client retried into a listener that was
+already gone. `AbstractTransportEngineTck$GracefulDrain` now pins the contract — it holds a stream
+mid-exchange across `stop()` and asserts the response still arrives.
+
+**Telemetry.** `CommunityTransportDrainEvent` (JFR `eu.exeris.kernel.transport.CommunityTransportDrain`)
+records each drain's stream count at start, count remaining, and duration. A non-zero
+`streamsRemaining` means the deadline fired before the drain finished — the signal that in-flight work
+was severed, and the number to tune `terminationGracePeriodSeconds` against.
 > `RST` is reserved for hard timeout scenarios only.
 >
 > **`reset(long)` vs `close()`:** `close()` is a graceful end-of-stream (`FIN`, drains queued writes);
--- a/docs/subsystems/transport.md
+++ b/docs/subsystems/transport.md
@@ -296,6 +296,14 @@ When PAQS sheds a stream or the Kernel initiates graceful shutdown:
 > **Why `FIN` not `RST` for load shedding?** `RST` causes immediate connection teardown on the client
 > side, which may interrupt in-flight retries and force the client to reconnect. `FIN` allows the client
 > to receive the `HTTP 503` response body, which is machine-readable and enables intelligent backoff.
+> `RST` is reserved for hard timeout scenarios only.
+>
+> **`reset(long)` vs `close()`:** `close()` is a graceful end-of-stream (`FIN`, drains queued writes);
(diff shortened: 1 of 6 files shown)


arkstack-dev:
Both findings addressed in `19193f7b`.

## Finding 1 — the flag-ordering window

Right, and it deserved to block: the entire premise of this PR is closing a shutdown-ordering bug, and I reintroduced the same class of it at instruction scale.

**The straightforward swap does not work, though.** Raising `draining` before the `running` CAS means a concurrent second `stop()` raises the flag, loses the CAS, returns — and leaves `draining` up permanently. The reactors of a subsequent `start()` could then never be allowed to exit. That trades a two-instruction window for a permanent leak.

Entering through a CAS on `draining` fixes both properties at once:

```java
if (!draining.compareAndSet(false, true)) {
    return;                       // a concurrent stop() loses the race here, atomically
}
if (!running.compareAndSet(true, false)) {
    draining.set(false);          // not running — do not leave the flag raised
    return;
}
```

The guard becomes atomic instead of two independent checks, and the invariant you named — at least one of `{running, draining}` reads true from the moment a stop is claimed until Phase 3 — now holds by construction. I wrote it onto `isReactorActive()`, since that predicate is what depends on it.

**Not converted to a tri-state**, which I agree is the cleaner end state. `running` is read by `isRunning()`, `setStreamHandler`, `setConnectionHandler`, `close()`, and `start()`'s failure path; moving all of them belongs in its own change rather than one whose subject is shutdown ordering.

## Finding 2 — the blockquote splice

Confirmed against the base branch and fixed. `...enables intelligent backoff.` and `` > `RST` is reserved for hard timeout scenarios only. `` were contiguous lines of one blockquote (`origin/development/0.11.0:docs/subsystems/transport.md:297-298`), and I inserted directly into that seam — splitting the quote and stranding the answer twenty lines from its question. You read the structure correctly.

The subsection now sits after the complete blockquote, before the `---`. `git diff origin/development/0.11.0 -- docs/subsystems/transport.md` is purely additive: no existing line is modified.

## Verification after the changes

- `mvn clean install` — BUILD SUCCESS, `MAVEN_EXIT=0`, full reactor, no skip flags
- `ExerisArchitectureTest` — 13/13
- `CommunityNativeTcpCarrierTckTest` — 10/10 including `GracefulDrain`
- Transport stress gate re-run — RUN+GREEN, 4 tests

Thanks for tracing `PaqsScheduler.close()` back to the `activeStreamCount()` spin-poll to confirm the fix's premise independently — that the drain can only make progress while something is still decrementing the counter is exactly why the old ordering made it inert, and it is the part a reader is most likely to take on trust.

PR #290, opened 2026-08-05
feat(ci): gate SPI compatibility on the declared stability matrix (ADR-065)

arkstack-dev:
Companion reservation: exeris-docs#50 (ADR-065 registered before this content was written).

## Why

`docs/stability-matrix.md` declares which SPI surfaces are settled. Nothing checked it. A contract could move on a surface labelled `stable` and nobody would learn about it until a consumer failed to compile.

Regenerating the record after the fact shows **three** binary-incompatible transitions between 0.5.0 and 0.10.2 — none announced as such at the time, and one of them (`TelemetryConfig.blackBoxOffHeapBytes()` → `glassBoxOffHeapBytes()`, v0.9.0) invisible to any review that reads diffs rather than bytecode, because the file survived and only a record component moved.

That is the sharper point: **a source-level diff is not a sufficient instrument for API compatibility.** It cannot see removed record components, removed constructors, or interface methods dropped where the file survives. Applied to 0.5.0 → 0.10.2 it reports one incompatible transition; bytecode comparison finds three.

## What lands

**Gate** (`tools/spi-api-diff/`, outside the reactor alongside `tools/jfr-reporter`) + a `spi-compatibility-gate` CI job. Binary-incompatible change to a `stable` surface fails the build; `preview`/`experimental` are reported, not gated — the declared semver policy executed rather than restated.

It compiles the SPI **from git** rather than resolving published artifacts. That is available only because The Wall constrains `exeris-kernel-spi` to `java.*`/`jdk.*`, which has a side effect nobody designed for: every revision of the module in the project's history compiles standalone with a bare JDK (verified across all ten release tags). So the gate needs no `PACKAGES_READ_TOKEN`, runs on fork PRs, and regenerates the whole release record from a clean clone.

**Docs** — `docs/release/spi-api-history.md` (generated per-transition record), `upgrade-0.5-to-0.10.md`, and reconstructed `v0.6.0-release-notes.md` (v0.6.0 shipped without notes and predates `CHANGELOG.md`, while carrying the only pre-declaration removal of `stable`-surface SPI types — the one release that needed notes was the one release without them).

**ADR-065** records the four rulings and the two findings above.

## What it found on its first run

- **`spi.scheduling` and `spi.storage.blob` had no matrix row** — shipping on this branch with accepted ADRs (057/056), `Abstract*Tck` coverage and Community bindings, and nothing declaring how much a consumer may lean on them. Both labelled `preview`.
- **`FlowSnapshot` (ADR-062) breaks `spi.flow`, declared `stable`.** Adding a component to a record changes its canonical constructor. **The gate fails on this PR's own branch, and that is the intended behaviour.** Taking the break pre-1.0 is defensible; taking it silently is what this ends. ADR-062 now records it as a decision owed at the v0.11 cut — retain the previous canonical constructor as an overload, or take it deliberately and say so in the notes. **Not decided here.**

## Drift also corrected

The matrix and support matrix asserted "no external SPI consumers", which conflated two claims. No consumer is under a support contract — true, and what the pre-1.0 caveat rests on. That integrations exist is a different statement. The documents now say the narrower thing they mean. ADR-061 and ADR-062 carried the same conflation and are ACCEPTED-but-unreleased on this branch, so they are live text and are corrected in place; released ADRs and shipped release notes are left alone, since they were accurate when written.

The record reads better after the correction, not worse: the matrix was first published in **v0.9.0**, and since then the declaration has been kept — the one breaking transition in that window lands entirely on `spi.events`, labelled `preview`, where the policy permits it.

## Verification

- Gate exercised on five revision pairs with known outcomes: `v0.5.0→v0.6.0` (fail, 7), `v0.8.1→v0.9.0` (fail, 1), `v0.9.0→v0.10.0` (**pass** — preview-only), `v0.10.2→HEAD` (fail, FlowSnapshot), `--verify-surfaces` (pass after classification).
- Two silent-pass defects were hit while building it and are now asserted against, because a gate reporting a false green converts an unknown into an assurance: japicmp separates includes with `;` (a comma-separated list matches nothing and reports clean), and an SPI revision that fails to compile yields an empty jar that compares as unchanged.
- No `.java` or `pom.xml` changes, so `mvn clean install`, lint and `ExerisArchitectureTest` are N/A — nothing in the reactor moved. `shellcheck` was unavailable locally; the script was verified behaviourally instead.


arkstack-dev:
## The gate was not broken — it was right, and the branch was behind

Branch updated from `development/0.11.0` (merge `1669e4ee`, no conflicts). The gate is green.

**What it had actually caught.** This branch's tip predated the `FlowSnapshot` decision. It carried ADR-062's `currentStepName` — which removes the eleven-parameter canonical constructor — without the compatibility overload that restores it. So the report was accurate:

```
---! REMOVED CONSTRUCTOR: PUBLIC(-) FlowSnapshot(long, long, String, int, FlowState,
     Instant, Instant, int[], int, byte[], long)
```

That is the 0.10.0 descriptor, and on that tree it was genuinely gone. The break was real; only the tree was stale.

Which makes this the strongest validation the deliverable could get: the gate flagged a real removal on a `stable` surface, and goes silent on the tree where the overload puts it back. Verified both ways rather than reasoned about —

| tree | result |
|---|---|
| branch tip `99a9ccf3` | `stable-breaks=1`, exit 1 — the removed constructor |
| tip merged with `development/0.11.0` | `stable-breaks=0`, exit 0 |

**Also confirmed on the merged tree:** `--verify-surfaces` reports *all 239 SPI classes at HEAD are classified*. The merge brought in new SPI types (`FlowMigrationState`, `FlowDefinitionMigration`, and the `spi.scheduling` / `spi.storage.blob` surfaces), and none escaped the config — which is the hole the per-class check was rewritten to close.

`preview-breaks=1` remains, and is `ImmutableStorageContext` dropping its five-argument constructor. Permitted on `preview`, and now called out in `docs/release/v0.11.0-release-notes.md` as the semver policy requires.

`mvn clean install` green on the merged branch — full reactor, no skip flags.

## One thing worth deciding before this merges

The gate has **no acceptance mechanism** — no waiver file, no baseline, no accepted-list; `--fail-on-stable` is binary. That is fine today at zero stable breaks, and building machinery for zero cases would be the wrong trade. But the first *deliberate* stable break has nowhere to be recorded, and a permanently red gate is one people learn to mute — after which the next un-accepted break hides in the noise.

Not proposing it for this PR. Worth writing the policy down before it is needed rather than inventing it under release pressure: a narrow per-class entry carrying surface, change kind, version, ADR and rationale, self-expiring once the comparison baseline passes the accepting version, and surfaced in the report as ACCEPTED rather than being indistinguishable from green.

PR #291, opened 2026-08-06
feat(flow): in-flight migration between definition versions (ADR-064 stage 2)

arkstack-dev:
Second half of ADR-064. Coexistence (#289) let two versions live side by side and made a parked saga resume on the version it recorded; it left no way to move that saga forward, so retiring a version meant keeping it registered forever. This adds the transform that moves it.

## What lands

**SPI** — `FlowMigrationState` (parked step index and name, compensation stack, stack pointer, opaque state) and `FlowDefinitionMigration`, a functional interface taking one and returning one. `FlowExecutionPlanFactory.registerMigration(definitionName, fromVersion, migration)`.

**Core** — on wake, a saga whose version is not hosted walks `v → v+1` through registered transforms until it reaches a version the engine hosts, and resumes there. Bounded at 32 hops as a blast radius, not as the termination mechanism.

**Compatibility** — the 0.10.0 `FlowSnapshot` canonical constructor is restored as an overload. See below; this is the item that most needs a reviewer's eye.

## Three things settled during implementation

Recorded as ADR-064 amendments rather than left as silent divergence from the accepted text.

**Wake only.** `schedule()` fixes the target version at the plan the caller supplies, which makes the chain's terminating condition path-dependent — one policy cannot serve both doors. It also keeps application code out of a `liveInstances.compute` mapping function, out of `lookupParked`'s read-only query, and away from a `COMPENSATING` snapshot reaching a transform written for a parked saga. A resubmit against a mismatched version keeps failing closed. Stated plainly as a functional narrowing: choreography can reach `schedule()` directly, and a saga resubmitted rather than woken is not migrated.

**Stop at the first hosted version.** `planCatalog` offers point-gets only and the one name-scoped query is a full `keySet()` scan; adding another to the resume success path would be a No Waste Compute regression. So the walk tests one key per hop and returns the moment it is present.

**A successful migration is persisted**, before the resumed step runs. The alternative re-runs the chain on every wake, which makes transform purity a load-bearing obligation no document states, and leaves the row asserting a version the saga no longer runs — the same class of lie as a step recorded by position.

## SPI stability — the item worth a reviewer's attention

`eu.exeris.kernel.spi.flow` is declared **stable since 0.5.0** in `docs/stability-matrix.md`. This milestone adds two components to the `FlowSnapshot` record — `currentStepName` (ADR-062) and `definitionVersion` (ADR-064) — taking the canonical constructor from eleven parameters to thirteen.

The 0.10.0 descriptor is restored as an overload, so code compiled against 0.10.0 still constructs snapshots. Both new components default to their fail-closed sentinels, so the bridge grants compilation and **not** a bypass: a snapshot built through it is refused on resume by exactly the guards those two ADRs added. `AbstractFlowDefinitionVersioningTck$StabilityCompatibility` asserts both halves, because a bridge satisfying only the first would quietly undo both decisions.

What the overload cannot restore is stated rather than glossed: adding a component changes the record's *component list*, which deconstruction patterns and reflection over `RecordComponent[]` observe. Irreducible for a record. The alternative was to not record the parked step's identity and the definition version at all, which is the corruption class both ADRs exist to close — so the trade was made knowingly.

This is recorded now, in `docs/release/v0.11.0-release-notes.md` and on the matrix row, rather than at the release cut: a source-level diff does not surface a component-list change, an ABI diff does, and a stability declaration is worth something only if a violation is answered deliberately instead of found late.

## Verification

`AbstractFlowDefinitionVersioningTck` grows from 7 to 15 cases across four nests. Green, then mutated — a green contract test proves nothing until something wrong makes it red:

| Mutation | Killed by |
|---|---|
| migration result not persisted | `migrationIsPersistedBeforeTheResumedStepRuns` |
| chain hops past a hosted version | `migrationStopsAtTheFirstHostedVersion` |
| transform never invoked | `throwingMigrationFailsClosed` |
| duplicate registration silently replaces | `duplicateMigrationRegistrationIsRefused` |

Each killed by its own case and by no other. Three of those four cases exist because the sweep found the originals passing for the wrong reason — the throwing case asserted `RuntimeException`, which `FlowEngineException` also is, and the persistence case could not separate the migration write from the park write that followed it.

Gates run:

- `mvn clean install` — full reactor, no skip flags, so lint-gated (`pmd:check` at `verify` executed).
- `ExerisArchitectureTest` — 13 rules green.
- `integration` gate on `exeris-kernel-community` — `CommunityJdbcFlowSnapshotStoreTckIT`, 12 tests against real Postgres. This is the gate that went red on #289.
- `continuity` gate on `exeris-kernel-community` — `CommunityRestartRecoveryIntegrationTest` green.

## Docs

`flow.md` carried a row and a blockquote saying version-aware routing was "planned but not yet available" and "MUST NOT be relied upon" — three rows below the ones describing the coexistence that shipped in #289. Both replaced. The versioning TCK was also missing from the coverage table since #289; added.

ADR-064 amendment A2 described a `to == from + 1` check that does not exist — `registerMigration` takes no target version, so adjacency is unrepresentable rather than validated. Corrected to what the code does.

## Carried, not closed

- A rejected migration is not miss-suppressed: it throws before reaching the parked-lookup miss counter, so a repeatedly-woken unmigratable saga re-reads the store each time.
- Compensation-stack validation checks indices, not step identity — the ADR-062 treatment applied to the stack rather than the cursor.

Both are named in the ROADMAP entry rather than left implicit.


arkstack-dev:
## Added after review of the two carried debts: a compensation-stack guard, and a correction to obligation 9

Auditing the debts this PR listed turned one of them into a defect in a claim the PR itself makes, so it is fixed here rather than deferred.

### What obligation 9 actually covered

The ADR says a migration's output is validated by ADR-062's identity check. That check reads `currentStep` and `currentStepName` — the cursor. `FlowMigrationState` has five components, and the one it never looked at is the compensation stack: the component that decides what a rollback undoes, and one a transform may rewrite.

### Why this is a rollback bug, not a resume bug

`runCompensationStep` resolves each entry with `plan.stepAt(entry)`, and `stepAt` is a bare `steps[stepIndex]` — **outside its own catch**. So a stale entry throws out of `runCompensations`, skipping both the remaining unwind and `finalizeFailedInstance`, which is what releases the idempotency guard, clears the lookup-miss tracking, and writes the terminal state. The saga is left mid-compensation with its guard held — and only after something else has already failed, i.e. at the furthest possible point from the code that caused it.

The hole predates ADR-064: a shrinking redeploy can leave the same shape, and the cursor guards do not see it (right index, right name, wrong stack). What ADR-064 changed is that moving a saga onto a changed definition is now the *sanctioned* path, which turns an operator mistake into ordinary operation.

### What landed

The live prefix of the stack — below `stackPointer`, since entries above it are dead — is validated against the target plan on every resume, refusing with a new `COMPENSATION_STACK_OUT_OF_RANGE` reason on the existing `EX-FLOW-7002 / SCHEMA_MISMATCH` phase, leaving the row intact like every other resume refusal.

This is the **bounds** half. An entry that indexes the plan but names a step the saga never ran is not detectable from indices — the same blind spot the cursor's index had before ADR-062. Closing it means the stack carries identities, i.e. a third `FlowSnapshot` component, so it lands as its own slice.

**Scheduled inside v0.11 deliberately.** The record's component list already changed this milestone and the release notes already say so. A third component adds nothing further to the stability ledger; deferring it to v0.12 would be a fresh change to a surface declared stable since 0.5.0.

### The other debt is withdrawn — it was not a defect

`loadSnapshot` consults the miss cache *before* reading the store. Suppressing a version refusal would therefore make the documented remedy — register the missing transform and wake again — silently do nothing. Not suppressing is correct. The residual cost is one store read per wake of an unmigratable saga, and the invalidation mechanism already half-exists: `compile()` clears the tracking via `clearLookupSuppressionAfterPlanCompile`, `registerMigration` does not.

### Verification

TCK 15 → 18 cases. Three new: the plain-resume refusal, the migration-transform refusal, and one asserting that entries *above* `stackPointer` are dead and do not refuse a sound saga — fail-closed is only correct when what it closes on is live.

| Mutation | Killed by |
|---|---|
| guard removed | `compensationStackOutsideThePlanIsRefused`, `migrationEmittingOutOfRangeCompensationStackIsRefused` |
| scan the whole array instead of the live prefix | `staleEntriesAboveTheStackPointerAreIgnored` |

Gates re-run after the change: `mvn clean install` full reactor, no skip flags; `ExerisArchitectureTest` 13 rules; `integration` (12 tests, real Postgres); `continuity`.

PMD `TooManyMethods` fired on `FlowEngineException` at the eleventh factory. Suppressed with justification rather than collapsed — one named factory per rawArgs layout is what keeps string literals out of throw sites, and removing public factories from a stable SPI class mid-milestone is the worse trade. Same disposition as `PersistenceStatement`, `GraphSession`, `KernelProviders`.

arkstack-dev:
## Review round 2 — four code fixes, one documentation correction, and one of my own tests was vacuous

Thanks — both blocking findings were real. #1 turned out to be a defect in the *claim*, not in the code, and the reason is worth stating because the suggested fix would have broken something the review didn't have in view.

### 1. `lookupParked` — finding accepted, fix declined, claim corrected

You're right that `directPlan == null` does not distinguish wake from lookup, and that A1's text asserts otherwise. That text is wrong twice over.

**Wrong on the facts.** `lookupParked`'s durable-store branch was never a read-only query. Before this PR it already built a `RuntimeFlowInstance`, registered it in `parkedInstances`, cleared miss tracking and emitted `WakeOnLoadFallbackEvent`. Calling it read-only described nothing that was true; I imported the phrase from the *in-memory* branch, which genuinely is.

**Wrong on the consequence, and this is why I'm not taking the suggested fix.** `FlowChoreographyBridge` wakes as `lookupParked(most, least).ifPresent(scheduler::wake)`. On a node that doesn't hold the instance in memory, that store fallback **is** the wake's restore step. Threading `allowMigration=false` into it wouldn't make it read-only — `resolveVersionedPlan` would then throw `DEFINITION_VERSION_UNRESOLVED` (it already does today for an unhosted version, that part isn't new), so a cross-engine saga on a retired version would **refuse instead of migrate**, on exactly the topology ADR-013 §8 defines the fallback for. That is a far larger narrowing than A1 claims to state plainly, and it would silently gut the feature for choreographed deployments.

So A1 is rewritten to say what the code does — migration is scoped to the **resume-restore** path, `schedule()` still refuses — and the two protections the original list *did* get right are kept written down, because they hold by construction: no transform runs inside a concurrent-map mapping function (the restore completes before `liveInstances.putIfAbsent`), and none sees a non-`PARKED` snapshot.

The residual you identified is real and now stated rather than denied: a bare introspection `lookupParked` can run a transform and write. It is bounded to **once** — A3's persistence is what stops a polled lookup re-running application code. Two new TCK cases, since you correctly noted nothing exercised `lookupParked` at all:

- `lookupParkedThenWakeMigratesAndResumes` — the choreography shape end to end.
- `repeatedLookupRunsTheTransformOnce` — three lookups, one transform invocation, row reads 2.

A genuinely read-only `lookupParked` is reachable — build a snapshot-backed context without binding a plan — but it changes what the call registers and what the cross-engine recovery IT observes, so it's a separate decision, recorded as such rather than smuggled in here.

### 2. Write-fence — fixed, and hoisted higher than suggested

Applying the fence at the save would migrate in memory and skip the write, which resumes the saga on a version its row does not name — the failure A3 exists to prevent. So the fence goes **above the walk**: a migration that cannot be persisted must not happen. The per-instance generation half doesn't apply (no instance yet; the restore binds to the current generation), so what remains is `started && !closed`.

`closedRuntimeDoesNotMigrate` pins it — and **my first version of that test was vacuous**. It fetched `engine.scheduler()` after `close()`, and `CoreFlowEngine.scheduler()` calls `ensureStarted()`, so the assertion was satisfied by that refusal and the fence was never reached; the mutation survived. Fixed to capture the scheduler *before* the close, which is also the shape real code has — `FlowChoreographyBridge` holds one from construction, and `lookupParked` has no started-check of its own. That's the abandoned worker the fence is for.

### 3–5, all taken

- `orElseThrow` → `isMigrationCandidate` declines a snapshot with no step identity, so ADR-062's own `STEP_IDENTITY_ABSENT` answers instead of a bare `NoSuchElementException`.
- `migrations` gains the plan catalog's ceiling — and needs it for the same reason rather than by symmetry: a transform deliberately outlives the version it moves from, so the map grows on the deploy cadence and the catalog's eviction can't trim it. Bound check and insert share one critical section, as `compile()` does.
- `FlowDefinitionMigratedEvent` — you're right that the success outcome was the invisible one. Definition metadata and instance identity only; `opaqueState` is user data and is never recorded.

### Sonar, from the same pass

- `FlowMigrationState` gains content-based `equals`/`hashCode`/`toString`. A record's generated `equals` compares array components by reference — an identity-sensitive operation on a carrier, and it would make the obvious way to assert a transform is a no-op quietly always false. `toString` prints `opaqueState`'s size, never its bytes.
- A Javadoc block I had inserted between an existing one and its method, orphaning it. Moved back.

### PMD fired twice afterwards, both accurately

`CoreFlowPlanFactory` went God-Class at `TCC=16.7%` — migration admission shares no state with plan compilation, so it moved to `CoreMigrationRegistry` rather than raising a threshold. `migrateIfNeeded` hit complexity 10; its preconditions extracted to `isMigrationCandidate`.

### Verification

TCK 18 → 22 cases. Mutations: removing the lifecycle fence kills only `closedRuntimeDoesNotMigrate`; removing the identity bail kills only `migrationOfSnapshotWithoutStepIdentityRefusesCleanly`.

Gates re-run at HEAD: `mvn clean install` full reactor with no skip flags (so lint-gated — it's what surfaced both PMD findings); `ExerisArchitectureTest` 13 rules; `integration` on real Postgres; `continuity`.

One note on your note: the first full build I ran this session reported success because I'd piped it to `tail`, so I was reading `tail`'s exit code. It was a `BUILD FAILURE`. Subsequent runs capture the exit code directly — worth knowing if you're reading my earlier gate claims.

arkstack-dev:
## Round 3 nit — fixed by changing the code, not the comment

Good catch, and an uncomfortable one: this is the same claim-vs-code gap the last two rounds were spent correcting, in code I had just written. I noticed while writing it that the version guard runs first, decided "either way it's a Glass-Box refusal", and then wrote a Javadoc asserting the reason that doesn't fire. That is exactly the reasoning the A1 correction exists to reject.

**Fixed in the direction you suggested as the better of the two.** Your remediation argument is what settles it: `DEFINITION_VERSION_UNRESOLVED`'s documented remedy is "deploy the missing version or register the missing transform", and on this path a transform is typically *already registered*. The row is unresumable because it records no step identity, which no deployment fixes. A refusal naming the wrong runbook is worse than a raw exception, because it looks actionable. So `migrateIfNeeded` now refuses explicitly rather than declining into the version guard, and the predicate splits: `needsMigration` stays a pure "is there a walk to do and may this runtime do it", `refuseRowThatCannotBeWalked` owns the refusal.

Scope kept narrow deliberately — this fires only for a row that *needs a walk* and cannot be walked. A pre-0.11 row missing both version and identity still refuses as `DEFINITION_VERSION_ABSENT`, which `legacyConstructorSurvivesAndStillFailsClosed` pins and which ADR-064 intends; both reasons there carry the same remedy, so there is nothing to disambiguate.

**You were also right that the test couldn't have caught it.** `rawArgs().hasSizeGreaterThan(2)` is satisfied by every `FlowEngineException` on this path. It now pins the reason string, and the mutation reproduces your hand-trace exactly:

```
expected: "STEP_IDENTITY_ABSENT"
 but was: "DEFINITION_VERSION_UNRESOLVED"
```

**Fence race note** — taken, and written into the Javadoc rather than left for the next person: the fence is read once before the walk, not again before the save, so a `close()` landing mid-walk is not caught. Same tolerance the surrounding code already has (`started`/`closed` are plain volatiles throughout this class), over a window of at most `MAX_MIGRATION_HOPS` in-memory transforms.

**`FlowSnapshot.toString`** — I took this one despite your marking it out of scope. It rendered `opaqueState` in full via `Arrays.toString`; that is the application's payload, the one component of the record that is user data rather than definition metadata, and `toString` is what reaches logs, debuggers and exception text. It now renders a byte count, matching `FlowMigrationState`. `compensationStack` stays rendered — step indices are definition metadata. The asymmetry was one this PR introduced by fixing the sibling carrier, and leaving two carriers holding the same bytes with opposite dispositions seemed worse than a one-line widening. No test asserted the previous output.

Gates re-run at HEAD: `mvn clean install` full reactor, no skip flags; `ExerisArchitectureTest` 13 rules; `integration` on real Postgres (12 tests); `continuity`. TCK 22 cases.

PR #292, opened 2026-08-06
fix(spi): give registerMigration a default, and record v0.11's preview break

arkstack-dev:
Fallout from actually running the SPI compatibility gate from #290 (`--old v0.10.2 --new HEAD --fail-on-stable`) rather than predicting what it would say.

## The gate is green — and that corrects something in the notes

`stable-breaks=0`, exit 0. `FlowSnapshot` gaining two record components is **not** a binary-incompatible change, because the retained 0.10.0 constructor means nothing was removed and the new components only add accessors. The release notes claimed the opposite ("a source-level diff does not surface a component-list change; an ABI diff does"). Corrected: the reflective and deconstruction-pattern exposure is real, but it is precisely the kind of change an automated binary gate is silent about.

## What running it did surface

**`FlowExecutionPlanFactory.registerMigration` was abstract on a stable interface.** ADR-064 added it to a package declared stable since 0.5.0 with no default, which compiles away every existing implementor. It now defaults to refusing with `UnsupportedOperationException`.

The default refuses rather than no-ops on purpose. A factory that cannot chain versions should say so; a registration silently swallowed surfaces much later as a saga refused with `DEFINITION_VERSION_UNRESOLVED`, whose remedy is "register the missing transform" — which the operator would already have done. Same rule and same disposition as `FlowSnapshotStore.listParked()`'s default in 0.7.0.

**The gate structurally cannot catch this.** Adding a method to an interface is binary-compatible: existing implementors link fine and only hit `AbstractMethodError` when it is invoked. japicmp is silent; the break is source-only. `FlowExecutionPlanFactorySourceCompatTest` closes it the only way available — by *containing* an implementor written against the 0.10.x shape, so removing the default stops the test source compiling. That is the assertion, and it was verified by removing the default:

```
PreV011Factory is not abstract and does not override abstract method
registerMigration(String,int,FlowDefinitionMigration)
```

**`spi.security`'s preview break was unrecorded.** `ImmutableStorageContext` gained a sixth component (`sharedScopeKey`) this milestone and did not retain its five-argument constructor, so it is binary-incompatible. `preview` permits that — but the semver policy in the stability matrix promises preview breaks are called out in release notes, and the notes carried only the `FlowSnapshot` section. Added, with the `withSharedScope()` composer named as the way to stay insulated while the carrier is still `preview`.

## Verification

- `mvn clean install` — full reactor, no skip flags, so lint-gated.
- `ExerisArchitectureTest` — 13 rules.
- `FlowExecutionPlanFactorySourceCompatTest` — 2 cases; the compile half proven by removing the default and watching the build fail.
- SPI gate re-run after the change stays at `stable-breaks=0`.

No behaviour change for any binding in this repo — `CoreFlowPlanFactory` already overrides `registerMigration`.


arkstack-dev:
## Both taken — and #1 is the sharper of the two

**1. The matrix held the falsified claim, and it is the doc that wins ties.** Fixed. I corrected the release notes and left the higher-precedence source of truth carrying the version this PR exists to disprove — in a doc that says of itself *"if they ever disagree, this table wins and the drifting doc is the bug"*. That is the correction being half-applied, which is worse than not applying it, because the tie-break rule now resolves in favour of the wrong fact.

The rewrite strengthens the row rather than softening it, which I think is the more useful framing: **no automated gate reports this at all.** japicmp finds nothing because nothing was removed; a source diff shows the components arriving but not that anything downstream depends on their number or order. So the row is the record — not a human note duplicating a machine that would have caught it anyway. The same asymmetry now explicitly covers `registerMigration`, since it lands on the same package in the same cut.

**2. "Same disposition" was wrong, and self-contradictory two paragraphs up.** You're right: `listParked()` returns `List.of()` — a silent degradation — while this one throws, and the same Javadoc had already said "the default refuses rather than no-ops on purpose". Same *rule*, deliberately different body.

Rather than just deleting the claim I've stated why they differ, because the divergence is load-bearing: an empty parked list is a truthful degenerate answer for a store that does not track them, whereas there is no truthful degenerate answer to "register this transform" — accepting one and never applying it reports success for work that will not happen. The Javadoc now says to read the two together as "a stable interface needs a default", never as "defaults no-op". Release notes carry the same clarification, since a reader arriving there shouldn't have to open the Javadoc to learn the precedent diverges.

## Verification

`mvn clean install` full reactor, no skip flags (so lint-gated); `ExerisArchitectureTest` 13 rules. Both green at `cca5b822`.

On your "not verified" note — the sandbox blocking `mvn` is worth flagging back: two of the last three rounds on #291 turned on things only a build surfaced (two accurate PMD findings, and a mutation that exposed one of my own tests as vacuous). A source-level review catching both of this round's items without a build is a good result, but it's the class of gap where the next miss is likelier to be silent.

Issue #282, opened 2026-08-05
Graceful drain waits for idle keep-alive connections — shutdown takes the full 60s deadline after any request


arkstack-dev:
Found while evaluating the 0.11.0 pin from `exeris-spring-runtime`. Reproduced with a **kernel-native handler and the kernel testkit only** — no Spring involved.

## Symptom

After a single completed request from a keep-alive client, `TransportEngine.stop()` does not complete until the PAQS drain's 60-second hard deadline expires. The listening socket keeps serving normally for that whole time.

## Measurements

`EmbeddedHttpEngineFixtures.kernelBootstrapFixture()` + a trivial handler:

```java
private static final HttpHandler ECHO = exchange ->
        exchange.respond(HttpResponse.noBody(HttpStatus.OK, exchange.request().version()));
```

| Scenario | `fixture.close()` duration |
|---|---|
| No request at all | **5 ms** |
| One GET, JDK `HttpClient` left open (keep-alive) | **10 013 ms** — the fixture's own `STOP_TIMEOUT_SECONDS` join deadline, after which `FixtureThreads.joinQuietly` interrupts and returns while the kernel is still draining |
| One GET, `client.close()` before shutdown | **2 ms** |

Measured end-to-end against a live socket, the port stays reachable and answers **200** for roughly **60 s** after shutdown begins — matching the documented deadline in `NativeTcpCarrier.stop()`: *"waits for the admission controller's active-stream count to reach zero under a 60-second hard deadline"*.

## Diagnosis

The drain appears to count an **idle keep-alive connection** as an active stream. The distinguishing evidence is the third row: the only difference is whether the client closed its side first. Nothing is in flight in either case — the response was fully received before `close()` was called.

## Why this matters more than the gap it fixed

0.10.2 dropped in-flight requests but shut down promptly. 0.11.0 preserves in-flight requests but will not terminate for ~60 s whenever a client holds a connection open — which is every load balancer, service mesh, and connection-pooling HTTP client in production. Kubernetes' default `terminationGracePeriodSeconds` is 30 s, so the normal outcome is SIGKILL mid-shutdown, on every rollout.

Put differently: the drain currently waits for **connections** to go away, but the property worth waiting for is **requests** finishing. An idle keep-alive connection will never go away on its own — that is what keep-alive means.

## Suggested direction (not a patch — this is your call)

[...]

PR #293, opened 2026-08-06
fix(transport): CAS the drain handle's flag, and make observing zero a commit

arkstack-dev:
Two defects in the graceful-drain machinery from #284, found reading the code rather than by a failure. Both are latent today and both destroy exactly what the class exists to guarantee.

## 1. The handle claimed thread confinement; the mechanism guarantees the opposite

`StreamWork`'s flag was a plain `boolean`, its Javadoc saying *"confined to the stream's own thread — the protocol loop that reports it is the only caller"*.

But the handle reaches protocol code **only** through `TransportScopes.STREAM_WORK`, a `ScopedValue` — `CommunityHttpRequestProcessor` never receives it as a parameter. And `PaqsScheduler` mandates, two screens above the binding site, that *"all concurrent operations within `runStream()` MUST use `StructuredTaskScope`"* — whose `fork()` inherits `ScopedValue` bindings. Any subtask written to that mandate can read the handle and call `markIdle()`/`markBusy()` from another thread. `ScopedValue` is chosen precisely for down-stack visibility and fork inheritance; a parameter or a field would have expressed confinement without the door.

Both races are fatal to the purpose:

| Interleaving | Result |
|---|---|
| two threads see `busy == true` in `markIdle()` | double release → count below the truth → drain ends early → **teardown severs a handler mid-response** |
| two threads see `busy == false` in `markBusy()` | double acquire → count never reaches zero → full 60s deadline, i.e. #282's symptom |

The flag is now compare-and-set, so idempotence is a property of the type rather than an assumption about callers. One uncontended CAS per transition on a per-stream handle — noise against the `getAndAdd` on the shared counter that already happens.

Ordering detail worth flagging for review: `markBusy()` takes the count **before** setting the flag. The reverse needs an optimistic set plus a revert on failure, which opens a window where a concurrent `markIdle()` observes `busy == true` and releases a count this handle never took.

## 2. Observing zero was not the same as committing to teardown

The loop was `while (drainCoordinator.busyStreams() > 0)`, and `markBusy()` did not consult `draining` at all. An idle keep-alive stream has `busy == false`; a request arriving after the loop's last read and before teardown flips 0 → 1 on an engine already tearing down.

The existing protection is cooperative — protocols read `isDraining()` and stop extending, and the H1 codec answers `Connection: close`. That covers a codec that checks. It does not cover a request already sitting in the socket buffer when the drain began, nor a protocol that does not read the flag. The same reasoning that makes *busy* the default applies one level up: a protocol that does not participate must not be able to silently re-arm.

**Checking the flag inside `markBusy()` would not have fixed it** — read `!sealed`, loop seals, increment lands after the commit. So the count and the commit move together: `sealIfIdle()` is a CAS to a sentinel, the drain loop spins on *that* rather than on a read, and `markBusy()` answers `false` once sealed. The H1 loop returns instead of serving — the honest reading of a connection whose peer was already told to close; RFC 9112 §9.6 requires a client to tolerate a persistent connection closing. On the deadline the coordinator seals unconditionally, since once shutdown proceeds regardless, letting a late arrival believe it is protected is worse than closing on it.

This was **not** a deliberate best-effort decision — the window was simply not considered when #284 was written. It is now stated in `transport.md` either way.

## Verification

`DrainCoordinatorTest` 11 → 19 cases: 5 on sealing, 3 on handle concurrency.

The concurrency cases were verified against the *previous* shape rather than assumed adequate — reverting the flag to a plain field with read-modify-write fails **all three**, the double-acquire case reporting a count of 3 against an expected 1. They cannot prove the absence of a race (that needs jcstress, not wired into this repo) and the test Javadoc says so rather than letting a green run read as proof.

- `mvn clean install` — full reactor, no skip flags, so lint-gated. It surfaced two real findings on the way (a Checkstyle declaration-order break and a PMD `UnnecessaryWarningSuppression`, the latter correct — the flag *is* read directly).
- `ExerisArchitectureTest` — 13 rules.
- `stress` gate on `exeris-kernel-community` — 4 tests.
- Drain coverage specifically: `CommunityNativeTcpCarrierTckTest` (incl. `GracefulDrain`) + `CommunityHttpDrainIntegrationTest`, 12 tests.

One flake to record rather than bury: the first full build failed `CommunityConnectionRefusalTest.refusalIsCountedAndRecorded` (empty JFR event list, 0.025 s). Re-run 3× in isolation and twice as a full build: green every time. It is also not mechanically reachable from this change — the recording is stopped *before* `engine.stop()`, so the drain is not on that assertion's path. Recorded because this test family has been flaky on CI before, not because it is attributed to this PR.

### DIFF
--- a/docs/release/v0.11.0-release-notes.md
+++ b/docs/release/v0.11.0-release-notes.md
@@ -65,3 +65,25 @@ binary-incompatible change: code compiled against 0.10.x that constructs the rec
 recompiled. Use the `withSharedScope()` composer rather than the canonical constructor where the axis
 is not being set — the shared-scope key is orthogonal to the isolation strategy, and composing keeps
 call sites insulated from further components on this carrier while it remains `preview`.
+
+## Benchmarking: v0.11 moves the HTTP/1.1 per-request hot path
+
+Numbers measured on 0.10.x and on 0.11.0 are **not comparable without annotation**, and this is stated
+before a measurement campaign rather than after one.
+
+`DrainCoordinator` did not exist when the pre-0.11 figures were taken. It adds two atomic
+read-modify-writes on a shared per-engine counter to the H1 keep-alive loop — `markIdle()` when the
+connection parks on the next request, `markBusy()` when one arrives — so the cost lands **per request**,
+not per connection, and on one cache line shared by every stream the engine is serving. Whether that is
+visible above a given harness's noise floor is a question for a measurement, not for release notes; that
+it changes the path is not.
+
+The concurrency shape settled during this milestone and the final form is what any campaign must
+measure. Both operations are unconditional `getAndAdd` — the seal that closes the drain's commit window
+is expressed as a **negative range** rather than an exact sentinel value precisely so the hot path does
+not need a compare-and-set retry loop, which is the form that degrades under contention. An interim
+version during review did use CAS loops; it is not what ships.
+
+Treat 0.11.0 as a fence in the benchmark record, in the same way individual commits that moved this path
+have been annotated before. A figure carried across it without a note will read as an unexplained
+regression to whoever compares them next — including the person who took them.
--- a/docs/subsystems/transport.md
+++ b/docs/subsystems/transport.md
@@ -330,9 +330,15 @@ being dropped by it. On the 60-second deadline the coordinator seals uncondition
 shutdown proceeds regardless, letting a late arrival believe it is protected is worse than closing on it.
 
 **The per-stream handle is not thread-confined.** `StreamWork` reaches protocol code through the
-`STREAM_WORK` `ScopedValue`, and `StructuredTaskScope.fork()` — mandated for concurrency inside
-`runStream()` — inherits `ScopedValue` bindings, so a subtask can report idleness from another thread.
-Both the shared count and the handle's flag are therefore compare-and-set. A double release drops the
+`STREAM_WORK` `ScopedValue` — chosen precisely so it survives down the stack and across a task
+boundary. The reachable route is track-independent and does not rest on `StructuredTaskScope` (which
+the default line must not ship for 1.0 GA): `StreamExecutionBackend`, the seam at which an admitted
+stream's root task is started, is contractually required to preserve `ScopedValue` bindings across
+whatever thread a backend chooses, and its forward consumers — an Enterprise locality-aware backend,
+the post-1.0 DST simulation scheduler — are exactly the ones that run the task on a thread the
+scheduler did not create. On the `preview` artifact `StructuredTaskScope.fork()`'s binding inheritance
+is a second route to the same place. Both the shared count and the handle's flag are therefore
+compare-and-set. A double release drops the
 count below the truth and ends the drain early; a double acquire means it never ends. Neither is
 reachable by anything short of a stress harness, so the guarantee is a property of the type rather than
 an assumption about callers.
--- a/docs/subsystems/transport.md
+++ b/docs/subsystems/transport.md
@@ -316,6 +316,27 @@ When PAQS sheds a stream or the Kernel initiates graceful shutdown:
    looping here, so a handler that completes during the drain can still have its response flushed.
 3. **Teardown** — reactors are woken and joined, then the selector and all remaining channels close.
 
+**Observing zero and committing to teardown are one step.** `DrainCoordinator.sealIfIdle()` is a
+compare-and-set on the busy count, not a read followed by a decision. Read as two steps, a request that
+arrived while the drain was looking flips the count back to one on an engine already past its commit
+point, and phase 3 severs the stream serving it — the failure phase ordering exists to prevent, reached
+from the other side. Asking protocols to check `isDraining()` first does not close it: a request already
+sitting in the socket buffer when the drain began has nobody left to ask.
+
+After the seal, `StreamWork.markBusy()` answers `false` and the protocol layer closes rather than
+serves — `CommunityHttpRequestProcessor` returns from its keep-alive loop. That is the honest reading of
+a connection the peer was already told to close: a request arriving after that is racing teardown, not
+being dropped by it. On the 60-second deadline the coordinator seals unconditionally, because once
+shutdown proceeds regardless, letting a late arrival believe it is protected is worse than closing on it.
+
+**The per-stream handle is not thread-confined.** `StreamWork` reaches protocol code through the
+`STREAM_WORK` `ScopedValue`, and `StructuredTaskScope.fork()` — mandated for concurrency inside
+`runStream()` — inherits `ScopedValue` bindings, so a subtask can report idleness from another thread.
+Both the shared count and the handle's flag are therefore compare-and-set. A double release drops the
+count below the truth and ends the drain early; a double acquire means it never ends. Neither is
+reachable by anything short of a stress harness, so the guarantee is a property of the type rather than
+an assumption about callers.
+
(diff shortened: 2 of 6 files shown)


arkstack-dev:
## Fixed — and you were right that the existing test could not have caught it

Real defect, mine, introduced in this PR. `release()` now mirrors `tryAcquire()`'s CAS loop and declines on the sentinel.

Your walk is exactly what happens, and the wrap reproduces on demand:

```
expected: 0
 but was: 2147483647
```

`MIN_VALUE - 1` → `MAX_VALUE`, no longer the sentinel, so `tryAcquire()` resumes handing out counts and a stream can report itself busy mid-teardown — the failure this PR set out to close, reopened through the path that exists for it. And deterministic, not an interleaving: every timed-out drain with work in flight.

**The invariant was overstated too**, which is probably why I did not see it. `StreamWork`'s Javadoc claimed the handle holds exactly `busy ? 1 : 0` counts, full stop. `sealNow()` deliberately breaks that — it discards counts of streams still in flight, so a handle can be `busy` while the owner holds nothing for it. Once that is written down, "so every counter operation from here on is a no-op, which is why `release()` must decline rather than subtract from a sentinel" follows immediately. An invariant stated without its exception is what let an unguarded decrement look symmetric with a guarded acquire.

**On the test gap** — `sealNowDiscardsInFlightWork` was worse than incomplete, it was vacuous. It asked the *already busy* handle whether it could re-arm, and `markBusy()` answers that from its own flag before the coordinator is ever consulted, so it passed against no seal at all. It now asks a fresh handle. The new case, `releaseAfterForcedSealDoesNotCorruptTheCount`, walks the real path — seal while busy, then `close()` — and it is the **only** case in the suite that kills the unguarded version, which is a fair measure of how much coverage that path had.

That is the fourth test I have written this session that passed for the wrong reason, and the first where a reviewer rather than a mutation caught it. Worth noting since the same shape keeps recurring: an assertion that short-circuits before reaching the thing it names.

## Verification at `647aa7cd`

- `DrainCoordinatorTest` 20 cases (6 sealing, 3 handle concurrency, 11 pre-existing).
- Mutation: reverting `release()` to the unguarded `getAndAdd(-1)` kills `releaseAfterForcedSealDoesNotCorruptTheCount` and nothing else.
- `mvn clean install` — full reactor, no skip flags, lint-gated.
- `ExerisArchitectureTest` — 13 rules.
- Drain coverage: `CommunityNativeTcpCarrierTckTest` (incl. `GracefulDrain`) + `CommunityHttpDrainIntegrationTest`, 12 tests.

Also confirming your read on `29e393f`: it is comment/docs only, and the `StreamExecutionBackend` route is what makes the confinement argument hold on the default line without `StructuredTaskScope` anywhere near it.

arkstack-dev:
## Third remark taken — and it changed the code, not just the record

The fence is right, and I'd add one thing to the costing: **it moved after you priced it.**

You costed the #284 shape — two `getAndAdd` on a shared int, i.e. `lock xadd`, wait-free, one round trip on the line. That is not what the seal fix left behind. `tryAcquire()` and `release()` had both become read-then-CAS loops, so per request the shared line took two `cmpxchg` with retry instead of two `xadd`, plus two per-handle CASes. Uncontended, a wash. Under contention it is strictly worse — `xadd` never retries and a CAS loop does, and one per-engine counter under a full connection ceiling is the contended case. Your "pod kontencją może być gorzej" had stopped being a hypothesis and become structural.

**Fixed by making the seal a range instead of a value.** Any negative count means sealed, written as `Integer.MIN_VALUE / 2` so the marker has ~1e9 of headroom in both directions:

- `release()` → unconditional `getAndAdd(-1)`. Post-seal decrements drive the marker further negative, still sealed.
- `tryAcquire()` → `getAndAdd(+1)`, check sign, revert if negative. One `xadd` on the common path; the revert is only reachable after teardown committed.
- `sealIfIdle()` keeps its CAS — drain loop, cold.

So the hot path is back to two wait-free `xadd`, with the same atomic seal. And the range removes the earlier defect by construction rather than guarding each mutator against it: the exact sentinel at `MIN_VALUE` sat one decrement from wrapping to `MAX_VALUE`, which is what the previous commit had to patch.

**One of the two new tests was vacuous when written** — the fifth this session. `sealSurvivesRepeatedPostSealReleases` registered streams *after* the seal, and a handle registered after a seal holds no count, so its `close()` is a no-op and it exercised nothing. It now registers 10,000 before the seal and releases them after. Both cases die on the no-headroom mutation; before the fix only one did, and for the wrong reason.

## Fence declared

`docs/release/v0.11.0-release-notes.md` gains a benchmarking section: 0.11.0 is a fence, `DrainCoordinator` did not exist when the pre-0.11 figures were taken, the cost lands **per request** not per connection, and the shape any campaign must measure is the `getAndAdd` one — the CAS-loop version existed only between two commits of this review and does not ship. Deliberately no numbers: whether it clears a given harness's noise floor is a measurement, and this repo is not where that measurement lives.

Left for `exeris-benchmarks`, not done here: the single-arm light-contract run on 0.10.x vs 0.11.0, and the annotation alongside the commits already marked in the sweep.

## Verification at `22bafe3e`

`DrainCoordinatorTest` 21 cases. Mutations: `SEALED = Integer.MIN_VALUE` (headroom removed) kills both range cases; `busyStreams()` comparing `== SEALED` instead of `< 0` kills the corruption case. Full reactor `mvn clean install` no skip flags, `ExerisArchitectureTest` 13 rules, drain coverage 12 tests across `CommunityNativeTcpCarrierTckTest` + `CommunityHttpDrainIntegrationTest`.

arkstack-dev:
## Both nits fixed — and a sweep found a third that was worse than either

You were right that these matter more than usual here: both defects in this PR started as a claim that outlived the code under it. So rather than fix the two you named, I grepped the changed files for every remaining `sentinel` / `MIN_VALUE` / `compare-and-set` reference. That turned up a third, and it was the sharpest of the three:

**`tryAcquire()`'s own Javadoc** read *"CAS loop rather than `getAndAdd`, because the sentinel must not be arithmetic on"* — sitting directly on top of a method that is now `getAndAdd`-then-inspect. A comment asserting the inverse of the code it documents, and precisely the one someone would read before "restoring" the CAS loop.

And a fourth, milder: the class-level allocation note still said *"both the shared count and the per-handle flag are compare-and-set."* That flattens the distinction the range was introduced to create. The flag is CAS; the count takes `getAndAdd` per request and CAS only in `sealIfIdle()`, once per drain.

All four corrected at `05e92a3d`:

| Site | Was | Now |
|---|---|---|
| `tryAcquire()` | "CAS loop rather than getAndAdd" | add-then-inspect, and why the range permits it |
| `StreamWork` invariant | "release() refuses to decrement" | decrements unconditionally; a sealed count has no value left to protect |
| class allocation note | "both … are compare-and-set" | flag CAS; count `getAndAdd` per request, CAS only in `sealIfIdle()` |
| test rationale | "SEALED is Integer.MIN_VALUE" | `MIN_VALUE / 2`, unguarded decrement is the design; describes what an exact marker would do instead |

On the test you flagged: keeping `releaseAfterForcedSealDoesNotCorruptTheCount` alongside the 10k-stream case is deliberate and now says so — it is the minimal case, so if the range property breaks it is what fails first. Both die on the no-headroom mutation.

`mvn clean install` green, full reactor, no skip flags. `DrainCoordinatorTest` 21 cases.

Thanks for the three passes on this one — the sentinel-wrap blocker in particular was a defect I introduced while fixing a different defect, and nothing in the suite would have caught it.

PR #294, opened 2026-08-06
docs(platform): state the two-track JDK policy as a rule, and record the JDK 28 confirmation

arkstack-dev:
## Why now

The two-track model was already decided and written down — but only inside ROADMAP §"Platform Baseline for 1.0 GA", while `CLAUDE.md`'s **Stack** line named a single JDK with no track distinction. That is the line a reader hits first, and it is why a track-blind claim got into a transport fix this milestone: `PaqsScheduler`'s class doc mandates `StructuredTaskScope` for concurrency inside `runStream()` — correct on the `preview` artifact, **disqualifying on the default line** — and I read it as simply true while justifying a concurrency fix in #293. That mandate has no code behind it either: the class carries no STS import, all five mentions were prose.

A policy that only exists in the section that owns it gets forgotten by everything that isn't that section.

## What lands

The rule, with the two selection criteria side by side:

| | Default line (`main`) | `preview` branch |
|---|---|---|
| JDK | **LTS only** — 25 today, 29 next | **Latest, LTS or not** — 28 today |
| Preview flags | none; `--enable-preview` is what is being removed | `--enable-preview`, by definition |
| Concurrency | virtual threads + explicit `ScopedValue` rebind (both GA) | `StructuredTaskScope` |
| Carriers | records / immutable finals, Valhalla-*ready* by discipline | JEP 401 value classes, once exercised |
| Artifact | the distributable `1.0` | `1.0-preview` |

The load-bearing sentence: **which track a statement is about changes what it means.**

And the part that was implicit: tracking the newest JDK on `preview` is a *requirement*, not an indulgence. That branch converges into `main` at an LTS, so if `StructuredTaskScope` exits preview in **29** the preview line *becomes* the GA line there — it has to have absorbed the API churn before the LTS lands, not after.

## JDK 28 confirmation

Checked 2026-08-04. Targeted JEPs: **401 Value Objects (Preview)**, **539 Strict Field Initialization in the JVM (Preview)**, 535 Shenandoah generational-by-default.

Both Valhalla-relevant JEPs are **preview**, so 28 does not move the GA floor — it is the preview branch's next JDK, exactly as the section's June analysis predicted. Recorded as confirmation rather than rewritten, so the record still shows when it was analysis and when it became fact.

**JEP 539 was absent from that analysis** and is worth naming: strict field initialization is the companion that makes value classes sound. So the "Valhalla-ready carriers" guardrail moves from a style rule to something **executable on the preview branch** once that branch is on 28 — which is the point of the branch existing. The guardrail now says which track it is enforceable on, and cites the two carriers that already carry hand-written value equality because a record's generated `equals` compares array components by reference (`FlowSnapshot`, `FlowMigrationState`).

## Scope

Docs only — `CLAUDE.md` + `docs/ROADMAP.md`, no code. No status change: the baseline is still **NOT STARTED**, and the default line still carries the same four real STS sites the section names.

`mvn clean install` green (full reactor, no skip flags) — run because `CLAUDE.md` changes are cheap to get wrong in a way that only a build notices, not because docs compile.


arkstack-dev:
## Both observations checked — one fixed, one already in flight

**"mid-migration" vs NOT STARTED — you're right, and it's fixed at `7ce351eb`.**

Not a register mismatch; an overstatement. The default line carries the *same four* STS sites it carried before the target was set, and no substitution has landed. "Is being made preview-clean" and "mid-migration" both imply motion there has not been.

That matters more in a guardrail doc than it would elsewhere, because it changes what a reader concludes from a fifth site appearing: "mid-migration" makes a fifth read as *slowing something down*, when the truth is it moves away from a line that has not moved at all. The bullet now states the target as a target, says NOT STARTED explicitly, and names the four files.

I also added **how to count them** — by import, not by mention — because I got exactly this wrong earlier in the milestone. Javadoc references inflate the figure to eight, and one of those eight is `PaqsScheduler` carrying an STS *mandate* in prose with no import behind it, which is precisely how a track-blind claim ended up load-bearing in a transport fix. Verified: all four named files carry exactly one such import each.

**`PaqsScheduler`'s javadoc — already fixed, in #293, not missing.**

It is corrected there (commit `29e393f`, the paragraph now reads track-by-track and points at `StreamExecutionBackend` instead of the STS mandate). This branch was cut from `development/0.11.0` **before** #293 merged, which is the only reason the old text is still visible here. Confirmed by grepping the #293 branch.

So no tracking item — the fix lives in the PR whose incident motivated it, which seemed the right home rather than folding a code-comment change into a docs-only PR. Merge order settles it either way; if #294 lands first the correction arrives with #293.

`mvn clean install` green, full reactor, no skip flags.

Thanks for verifying the two claims that could have drifted rather than taking the diff at its word — the carrier-equality one in particular was written from memory and is the kind I'd want checked.

PR #312, opened 2026-08-11
ci(actions): harden and modernise the GitHub Actions surface

arkstack-dev:
The v0.11.0 release PR's Dependency Review gate went red, and the cause was upstream of the release: `SonarSource/sonarqube-scan-action@v5` resolves to a revision covered by two high-severity advisories (GHSA-f79p-9c5r-xg88, GHSA-5xq9-5g24-4g6f), both cleared only from 6.0.0.

## Where the scan runs matters more than the pin

That action entered the tree through a PR into `development/0.11.0`. Dependency Review only triggered on PRs into `main`, so it went unscanned for 87 commits and surfaced at the cut — on the one PR nobody wants to reopen. It now runs on `development/**` too. The pin is the symptom; the trigger was the defect.

Exposure was bounded, for the record: the step passes no `with:` arguments and is gated on an absent `SONAR_TOKEN`, so it never actually ran. That bounds the risk, it does not make the gate wrong to fail.

## Versions

Eleven distinct action refs, several multiple majors behind: checkout v4 → v7.0.1, cache v4 → v6.1.0, upload-artifact v4 → v7.0.1, download-artifact v4 → v8.0.1, dependency-review v4 → v5.0.0, setup-java → v5.7.0, codeql-action → v4.37.6, sonarqube-scan-action → v8.2.1, claude-code-action → v1, actions-gh-pages → v4.0.0.

Inputs were checked against each new `action.yml`, not assumed: `cache-hit` still exists on cache v6, `distribution: jdkfile` + `jdkFile` still exist on setup-java v5.7, and the artifact actions still take the input set this repo uses.

`dependency-review` v5.0.0 is the interesting one. The deprecation issue announced `deny-licenses` removal for 5.x; the release turned out to be a node24 runtime bump that still declares the input. The comment in the workflow now records what the release contains rather than what the issue predicted.

## Pins that stay current

Everything is pinned by commit SHA with the version in a trailing comment, so a moved tag cannot change what runs. SHA pins do not decay visibly, though — the reason this repo drifted three majors is that nothing said so. `.github/dependabot.yml` is added for the actions ecosystem, which rewrites the SHA and its comment together. Maven is deliberately excluded: dependency bumps belong on the milestone path where the tagged gates run.

## A mandate that outlived the milestone

The Copilot instructions, both prompt packs, both implementer agent definitions, and the two `.github/skills` mirrors still told an AI reviewer to prefer `StructuredTaskScope` for orchestration concurrency — which ADR-066 removed from the default line in this release. The review workflow's own prompt still described the repo as JDK 26 with `--enable-preview`.

The `.claude/skills` copies had already been truthed up; the `.github` mirrors had not, so the guidance depended on which surface an agent happened to read. All of them now name `core.concurrent.StructuredScope`, scope `StructuredTaskScope` to the `preview` branch, and the review prompt states both tracks up front — including why two sites run in-thread rather than fork.

## Also

The `eu.exeris/*` modules are allow-listed for the licence check. They report an unknown licence because the dependency graph reads each module POM without resolving the parent that carries `<licenses>` — POM inheritance, not an unlicensed artifact.

## Verification

Workflows and `dependabot.yml` parse as YAML. No reactor sources touched, so the Java gates are unaffected — but CI on this PR now exercises every bumped action end to end, which is the actual proof.

### DIFF
--- a/.claude/agents/exeris-implementer.md
+++ b/.claude/agents/exeris-implementer.md
@@ -12,7 +12,7 @@ Delivery agent for writing and refactoring code without re-litigating architectu
 
 ## Primary Responsibilities
 - Implement requested behavior with minimal, targeted changes.
-- Apply Java 26+ runtime idioms where relevant (`ScopedValue`, `StructuredTaskScope`, FFM, immutable carriers).
+- Apply current runtime idioms where relevant (`ScopedValue`, FFM, immutable carriers, and structured concurrency through `core.concurrent.StructuredScope` — `StructuredTaskScope` is preview and belongs only on the `preview` branch).
 - Preserve The Wall and existing module boundaries.
 - Surface risks early when requested change conflicts with architecture constraints.
 
--- a/.claude/commands/community-implementer.md
+++ b/.claude/commands/community-implementer.md
@@ -8,7 +8,7 @@ Implement this as an Exeris Community/Open-Core change.
 Constraints:
 - Do not re-litigate architecture unless a direct violation is detected.
 - Preserve existing module boundaries.
-- Prefer explicit construction, ScopedValue, StructuredTaskScope, immutable carriers, and zero-copy/off-heap-safe patterns where relevant.
+- Prefer explicit construction, ScopedValue, structured concurrency (`core.concurrent.StructuredScope` on the preview-clean default line; `StructuredTaskScope` only on the `preview` branch), immutable carriers, and zero-copy/off-heap-safe patterns where relevant.
 - Avoid framework DI, ThreadLocal for runtime context, and unstructured orchestration in runtime paths.
 - Keep changes minimal and targeted.
 - If the change affects SPI-observable behavior, explicitly mark that TCK review is required.
--- a/.claude/skills/exeris-performance-contract/SKILL.md
+++ b/.claude/skills/exeris-performance-contract/SKILL.md
@@ -57,7 +57,7 @@ This skill validates that changes respect:
 7. **Modern replacement guidance (mandatory recommendations)**
    For every violation, suggest nearest compliant replacement:
    - `ScopedValue` instead of `ThreadLocal`
-   - `StructuredTaskScope` instead of unstructured async primitives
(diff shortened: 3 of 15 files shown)


arkstack-dev (commented):
## Exeris Kernel — PR Review
**Triage:** LOCAL · CI hardening and AI-tooling doc sync; no Java sources, no SPI or Wall boundary touched.

**Verdict: APPROVE** _(comment — self-review restriction prevents a formal approval vote)_

Diff is clean at development-tier review depth. No banned constructs, no lifecycle or Wall issues to flag — this is entirely `.github/` and `.claude/` surface.

Quick scan notes:

- `dependency-review.yml`: extending the trigger from `main`-only to `development/**` is the right fix — dependencies now get scanned at the point of entry, not at the release cut. The `allow-dependencies-licenses` entries for `eu.exeris/*` (POM inheritance, not unlicensed) and `SonarSource/sonarqube-scan-action` (runner infrastructure, not distributed code) are well-justified in the inline comments.
- `dependabot.yml`: grouping `actions/*` and `github/*` together but leaving third-party actions ungrouped is the correct tradeoff — third-party changelog reads should not be batch-skipped.
- All eleven action refs are now SHA-pinned with a `# vX.Y.Z` trailing comment. The compatibility checks cited in the PR description (setup-java `jdkFile`/`distribution: jdkfile`, cache `cache-hit`, artifact input shapes) are exactly what needed to be verified before a multi-major bump; the description confirms they were.
- AI tooling docs (`.claude/`, `.github/agents/`, `.github/prompts/`, `.github/skills/`, review workflow prompt): the `StructuredTaskScope` → `core.concurrent.StructuredScope` + track-aware language lands consistently across all nine affected files. The `InMemoryEventBus.publishAndAwait` / `SubsystemOrchestrator` in-thread carve-outs in the review prompt are a useful precision addition.

**Strong pattern:** SHA-pinning paired with Dependabot configured to rewrite both the hash and its comment together closes the loop neatly — the pin is immune to a moved tag, and Dependabot keeps the comment telling you which tag it is.

Pre-merge review. Release-merge gate runs separately on PRs into main.
False positive? Reply @claude false-positive \<id\> or thumbs-down.

---

PR #314, opened 2026-08-11
fix(security): streaming routes were not authorized, and search_path outlived its tenant

arkstack-dev:
Part of the v0.11 release-review fix sweep. Two findings, both in the security-relevant path.

## ADR-061 never reached a streaming route

`CommunityHttpRequestProcessor` resolves the stream route and dispatches it **before** the request dispatcher is called at all — and the route policy and the `SecurityInterceptor` both live inside that dispatcher. Every SSE route was therefore served with no requirement checked and **no `PRINCIPAL_CONTEXT` bound**: a route declared as requiring a scope returned 200 to an anonymous caller, and a handler that read the principal saw nothing.

What makes this one worth reading twice: **the decision itself was fully covered by `AbstractHttpRoutePolicyTck` the whole time.** Nothing tested that the streaming branch reached it. Complete coverage of a decision says nothing about the paths that never ask.

The authorization is now one method — `authorize()` — used by both `dispatch()` and a new `dispatchStream()`. They stay separate entry points because `dispatch()` ends with a "nobody responded, send 500" tail that would trample a stream writing its own head; but the *decision* must not be two implementations. Since the stream dispatcher runs the emit loop on the calling thread, the principal binding now holds for the stream's whole life rather than just its open.

## `search_path` outlived its tenant

`RlsConnectionInterceptor` issued `SET search_path` only on the `SEPARATED_SCHEMA` branch. It is **session-scoped**, and `persistence.perTenantPooling` defaults to `false`, so `CommunityTenantPoolRegistry` hands every non-`DEDICATED` tenant the same pool: a connection pointed at one tenant's schema goes on to serve the next request unchanged. A `SHARED` request inheriting it still has its `tenant_id` enforced by RLS — this is not a row-visibility hole — but it resolves unqualified names **in another tenant's schema**.

`SHARED` and `DEDICATED` now `RESET` it. `RESET` rather than a literal `public`, because the inverse of a session override is the session default, which a deployment may set per role or per database.

The class already carried this exact argument in `injectSharedScope`'s javadoc (*"Absence must be published as `""`, not left unpublished"*); it simply had not been applied to `search_path`.

## Verification

Both fixes are **mutation-proven**, not merely green:

- Reverting the processor to the shipped dispatch order fails `protectedStreamRouteWithoutAuthorizationReturnsUnauthorized` **and nothing else**.
- Dropping the `SHARED` reset fails `sharedStrategyResetsSearchPath` **and nothing else**.

`mvn clean install` green; community 1522 tests; lint in cycle (no skip flags); `ExerisArchitectureTest` green.

## Owed test debt

A cross-tenant case for the `search_path` reset in `CommunityIsolationStrategyEndToEndIT` (`@Tag("integration")`, Testcontainers Postgres) — the unit test proves the statement is issued; only the IT proves the leak is closed end to end. Tracked, not blocking: the unit coverage is mutation-proven and the tagged gate does not run in the default build.

## Notes for review

- Touches `CommunityHttpRequestProcessor.java`, shared with #315 (transport). Verified to auto-merge cleanly in either order — different hunks — but whichever lands second should re-run the community suite rather than trust the merge.
- No SPI change. Docs: `NO_DOC_IMPACT` (ADR-061 already documents the intended behaviour; this makes the code match it).

### DIFF
--- a/docs/subsystems/http.md
+++ b/docs/subsystems/http.md
@@ -215,6 +215,20 @@ The policy answers `RouteRequirement` (`permitAll` / `authenticated` / `requirin
 `PrincipalContext` into admit / `401` / `403`. Roles are not expressible here — see
 [`security.md`](security.md) for why the edge checks scopes and `@RequiresRole` stays at the method.
 
+**A stream open passes the same gate as a request, through the same code.** Opening an SSE stream is
+a request that happens to be answered by an engine rather than an exchange, so it is subject to the
+route requirement identically: `CommunityHttpRequestDispatcher.dispatchStream` runs the same
+`authorize` as `dispatch`, and only the response tail differs, because a stream writes its own head
+and must not receive the respond-once tail. Until 0.11 it did not: the streaming branch returned
+before `dispatch` was ever reached, so a bound policy and the `SecurityInterceptor` were both skipped
+and an SSE handler ran with no `PrincipalContext` bound. The separate entry point exists to keep the
+response handling apart, **not** the decision — one implementation, one place it can drift.
+
+The binding covers the stream's whole life rather than its open alone: the stream dispatcher runs the
+handler's emit loop on the calling thread, inside `SecurityInterceptor.intercept`'s scope. The denial
+exchange is supplied lazily, so an admitted open — every open, in a healthy deployment — does not
+allocate one.
+
 ---
 
 ### HTTP/2 stream admission (since v0.8 Sprint 5, HTTP-112)
--- a/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/http/CommunityHttpRequestDispatcher.java
+++ b/exeris-kernel-community/src/main/java/eu/exeris/kernel/community/http/CommunityHttpRequestDispatcher.java
@@ -34,6 +34,7 @@ import java.lang.foreign.MemorySegment;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Supplier;
 
 // CyclomaticComplexity: route dispatch table (auth, health, user handlers) — each branch is a
 // terminal decision, not reducible without introducing opaque indirection.
@@ -81,7 +82,7 @@ final class CommunityHttpRequestDispatcher {
             }
         }
 
-        if (!authorize(request, exchange,
+        if (!authorize(request, () -> exchange,
                 () -> handleWithinRequestSession(method, request, exchange, handler))) {
             return;
         }
@@ -107,11 +108,15 @@ final class CommunityHttpRequestDispatcher {
      * {@link SecurityInterceptor#intercept}'s scope.
      *
      * @param request    the parsed request that opened the stream
-     * @param exchange   used only to write a denial; an admitted stream responds through its engine
+     * @param denial     supplies an exchange used ONLY to write a denial. A supplier rather than an
+     *                   exchange because the admitted case — the common one — never needs it, and
+     *                   building one per stream open would spend an allocation on the path whose
+     *                   whole point is that it hands the socket to the stream engine instead
      * @param openStream opens the stream and runs its emit loop
      */
-    /* default */ void dispatchStream(HttpRequest request, HttpExchange exchange, Runnable openStream) {
-        authorize(request, exchange, openStream);
+    /* default */ void dispatchStream(HttpRequest request, Supplier<HttpExchange> denial,
+                                      Runnable openStream) {
+        authorize(request, denial, openStream);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


arkstack-dev:
Marking this draft: the Persistence RLS/Interceptor gate fails on `CommunityPersistenceSharedScopeIT.assert_write_stays_pinned_to_owner` with `new row violates row-level security policy for table "scoped_docs"`.

That gate is `@Tag("integration")` and does not run in the default build. It covers `RlsConnectionInterceptor` directly, so it was mine to run before opening this — I did not, and the "lint in cycle / community suite green" claim in the commit message says nothing about it.

The F2 half (stream-route authorization) is unaffected and independently verified. The F4 half (`RESET search_path` on SHARED and DEDICATED) is the suspect: the policy's `WITH CHECK (tenant_id = current_setting('exeris.tenant_id', true))` is what rejects the insert, and the only change on that path is the added reset statement. `set_config(..., false)` is session-level, so the obvious transaction-local explanation does not hold — needs a real diagnosis against the fixture rather than a guess.

Not merging until that is understood; splitting F2 out is likely the right shape.

arkstack-dev:
## Diagnosed — the root cause is pre-existing, and this branch is only the trigger

The stack trace settles it. The failure is **not** in `resetSearchPath`, which never runs:

```
RlsConnectionInterceptor.injectTenantId(RlsConnectionInterceptor.java:164)
RlsConnectionInterceptor.onConnectionAcquired(...:138)
CommunityPersistenceEngine.applyConnectionInterceptors(...)
CommunityPersistenceSharedScopeIT.readVisible(...:110)
Caused by: PSQLException: current transaction is aborted, commands ignored until end of transaction block
Caused by: PSQLException: new row violates row-level security policy for table "scoped_docs"
```

It dies on the **next acquire**, in unchanged code, on the interceptor's first statement.

### The defect

`JdbcPersistenceConnection` runs on a pool baseline of `autoCommit=false` (`CommunityHikariSupport:82`). That makes every statement open a real Postgres transaction even when the SPI caller never called `beginTransaction()` — which is exactly why `commitStandaloneWriteIfNeeded()` exists:

```java
private void commitStandaloneWriteIfNeeded() throws SQLException {
    if (inTransaction) { return; }
    if (!conn.getAutoCommit()) { conn.commit(); }
}
```

**That method has no counterpart on the failure path.** `close()` rolls back only `if (inTransaction)`, and `inTransaction` is set solely by an explicit `beginTransaction()`:

```java
if (inTransaction) {
    try { conn.rollback(); } catch (SQLException _) { }
    inTransaction = false;
}
```

So a **standalone write that throws** — which is what an RLS `WITH CHECK` rejection is, i.e. the security control working as designed — gets neither the commit nor a rollback. The physical connection returns to the pool inside a failed transaction block, and the next request to receive it fails on its first statement, whatever that statement is.

### Why it surfaced here

`assert_write_stays_pinned_to_owner` is the only case that deliberately provokes a rejected write and then immediately opens another connection. Adding `RESET search_path` to the SHARED acquire path changed the pool's assignment so step 2 receives the poisoned connection. **It was passing by luck of which connection came back.**

### Consequences beyond this test

Any RLS-rejected write poisons a pooled connection for whoever gets it next. In a deployment that is a cross-request failure with no relation to the request that caused it.

### Fix shape

Make the failure path symmetric with `commitStandaloneWriteIfNeeded`: in `close()`, roll back whenever the physical connection is not in autocommit, rather than only when `inTransaction` is set. Separate defect, separate branch — it is not F2 or F4 and should not ride this PR.

### What this means for #314

F4 (`RESET search_path`) looks correct; the gate failure is a pre-existing bug it exposed. This stays draft until the connection-return defect lands, then rebases onto it. F2 (stream-route authorization) is independent and mutation-proven — splitting it out to merge now is still the right call if the cut cannot wait.

arkstack-dev:
## Confirmed — diagnosis proven, and F4 was correct

Rebased onto `fix/v011-pooled-connection-aborted-transaction` (#322) and re-ran the gate locally with Docker:

```
CommunityPersistenceSharedScopeIT            4/4  ✓   (was 3/4, 1 error)
CommunityPersistenceFailedWriteReturnIT      2/2  ✓
CommunityPersistenceIsolationLeakTckIT       3/3  ✓
CommunityPersistenceTenantIsolationIT        2/2  ✓
CommunityIsolationStrategyEndToEndIT         3/3  ✓
                                            14 total, 0 failures
```

Command: `mvn -pl exeris-kernel-community -am -DincludedGroups=integration -DexcludedGroups= test`

So the `RESET search_path` change in this PR is sound; it only changed pool assignment order enough to expose #322's defect, which had been latent behind a 24-connection pool.

**Stays draft deliberately.** This branch now carries #322's commit in its history, and presenting it as independently mergeable would be stacking on an unmerged branch. Merge #322 first, then this rebases to its own two commits.

Also note the cross-tenant `search_path` case I listed as owed test debt: `CommunityIsolationStrategyEndToEndIT` already covers that ground and is green here. The debt was running the gate, not writing a case.

PR #317, opened 2026-08-11
fix(events): an Error escaping a handler leaked the payload's refcount

arkstack-dev:
> **Draft on purpose.** The regression test is owed and not written — see "Blocking" below. Opened now so the finding and the fix are visible alongside the rest of the sweep.

Part of the v0.11 release-review fix sweep.

## What was wrong

`InMemoryEventBus.publishAndAwait` caught `InterruptedException` and nothing else. The per-handler try-with-resources inside the dispatch loop releases the wrapper of the handler that threw, and `RuntimeException` is collected per handler — but an **`Error`**, or a failure closing one wrapper, unwound past the loop and left **every wrapper not yet reached open**. Those hold a `retain()` on the payload's off-heap buffer, so the refcount never returned to zero and the segment was pinned for the process's life.

The fork-per-handler shape this replaced in ADR-066 had the release in each subtask, so **the in-thread rewrite is where the gap appeared**: one collector for many wrappers, and only one exception type routed to it. Another instance of the milestone's recurring shape — a correct architectural change that moved a responsibility and dropped one path.

The fix releases the unclosed wrappers before rethrowing, with an explicit `//NOPMD AvoidCatchingThrowable` and the reason inline (release before rethrow, SPI boundary).

## Blocking: the regression test

**Not covered by a test yet.** The default suite passes (`InMemoryEventBusTest` 6/6, core green), which proves **nothing** about the leak — the assertion has to observe the payload refcount after an `Error` unwind, and belongs with the `LeakDetectionMode.PARANOID` fixtures rather than the behavioural ones.

This PR should not merge until that test exists and fails without the fix. The other five PRs in the sweep are independent of it.

## Notes for review

- One file, Core only. No SPI change, no boundary change.
- Docs: `NO_DOC_IMPACT`.

### DIFF
--- a/docs/subsystems/events.md
+++ b/docs/subsystems/events.md
@@ -482,6 +482,12 @@ operational differences relevant when targeting Kafka vs. Redpanda:
   > **(TCK gap — not yet implemented for the in-memory bus; the Kafka driver inherits per-key partition ordering from the broker by routing on `streamId` as the message key.)**
 - **Zero Subscriber Fast-Free:** Publish to a bus with zero subscribers; verify `payload.refCount() == 0`
   and slab is immediately returned to the pool (no silent leak).
+- **Broken Fan-Out Balance:** Fail the dispatch part-way — a `retain()` that throws, or a handler thread
+  that cannot be started — and verify the refs no handler will ever own are released before the failure
+  reaches the caller. Both `publish` and `publishAndAwait` owe this: the retain protocol above balances
+  only when every subscriber is actually reached, and the paths that do not reach them are the ones
+  where a leak is permanent. *(Closed for the in-memory bus by `InMemoryEventBusTest`; the retain
+  protocol is Core-local, so there is no `Abstract*Tck` anchor to inherit it.)*
 
 ---
 
--- a/exeris-kernel-core/src/main/java/eu/exeris/kernel/core/events/InMemoryEventBus.java
+++ b/exeris-kernel-core/src/main/java/eu/exeris/kernel/core/events/InMemoryEventBus.java
@@ -171,7 +171,13 @@ public final class InMemoryEventBus implements EventBus {
      * <p>Failure handling is unchanged — every handler runs, failures are collected, and the first
      * is thrown once all have finished.
      */
+    // java:S1181 — the same exemption the five other release-before-rethrow sites carry (see
+    // NativeCipherContext, SecurityInterceptor, PaqsScheduler). The Throwable is not handled here:
+    // it is rethrown unchanged, and the catch exists only so the wrappers no handler will ever
+    // reach are released first. Narrowing it would restore the leak for exactly the types that
+    // reach this path — an Error out of a handler is the one that motivated the fix.
     @Override
+    @SuppressWarnings("java:S1181")
     public void publishAndAwait(EventDescriptor descriptor, EventPayload payload)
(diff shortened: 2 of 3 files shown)


arkstack-dev:
## Ready — the owed test exists and is mutation-proven

`errorEscapingHandlerDoesNotLeakPayloadRefcount`: N=3, so the bus retains twice and owes three closes. **With the fix reverted the test records one** — the throwing handler's own try-with-resources — and the two wrappers no handler ever reached stay open, each holding a retain on the payload's off-heap segment for the process's life.

Two choices worth flagging to a reviewer:

- **It counts retain/close on a fake payload instead of allocating under `PARANOID` leak detection**, which is what I originally said it would need. The claim is about the refcount *balance* across N wrappers; a counting payload states that directly, where a `LeakDetectedError` reports the same fact one indirection away and only for payloads backed by a `LoanedBuffer`. The neighbouring `broadcastRaiiProtocol` case already uses this instrument for the same accounting.
- **The handler throws a `StackOverflowError`, not a `RuntimeException`.** The dispatch loop collects `RuntimeException` per handler, so it was precisely the types it did *not* collect that escaped past it. The test also asserts the Error still reaches the publisher — releasing the wrappers must not become swallowing what caused the unwind.

## SonarCloud

The failing check is `java:S1181` on the `catch (Throwable)`. Suppressed at `publishAndAwait` with the reason inline — the same exemption the five other release-before-rethrow sites already carry (`NativeCipherContext:214`, `SecurityInterceptor:153`/`:286`, `PaqsScheduler:333`/`:366`). The Throwable is not handled here; it is rethrown unchanged, and the catch exists only so the unreachable wrappers are released first.

`mvn -pl exeris-kernel-core -am clean install`, no skip flags — core 1236, lint in that cycle. `ExerisArchitectureTest` green.

arkstack-dev:
## Confirmed — the twin leak in `publish()` is real, and this PR does not fix it

Verified in source, five lines above the fix:

```java
if (slotCount > MULTI_SUBSCRIBER_THRESHOLD) {
    for (int i = 1; i < slotCount; i++) { payload.retain(); }   // N-1 retains
}
for (Slot slot : slots) {
    Thread.ofVirtual().start(() -> slot.handler().handle(descriptor, slotPayload));
}
```

If `start()` throws on slot *i*, slots *i..N-1* never receive a handler and their retains are never balanced — **exactly the "wrappers no handler will ever reach" this PR's own comment describes**, in the sibling method. There is no `TrackingWrapper` and no catch on this path at all.

Two halves, and they are not the same kind of problem:

- **A failed `start()` is unambiguously the bus's leak.** The retains were taken by the bus; nothing else can balance them.
- **A handler that throws before `close()`** is the documented RAII protocol being broken by the subscriber (the javadoc puts the close on the handler, and `broadcastRaiiProtocol` shows the `try (p)` shape). Whether the bus should defend against that is a contract decision, not a bug fix — and it would change `publish()`'s ownership model.

Not fixing it in this PR. The scope here is `publishAndAwait`, the fix and its test are complete and mutation-proven for that method, and widening now would mean shipping an ownership change to `publish()` without the TCK work that decision deserves. Tracked as a follow-up; if you would rather this PR cover the `start()` half only, say so — that part is a small wrap of the spawn loop and does not touch the protocol.

